PR #21, opened 2016-04-15
POT update and Brazilian Portuguese translation.

dvlopes:
Hi.

I've been working with undestrap in a few websites and translated it to Brazilian Portuguese (pt-BR).

Thanks.


holger1411:
Many thx @dvlopes !

PR #45, opened 2016-08-01
Adding source maps

akkolad:
it need to install the "gulp-sourcemaps" package

source-maps are usefull !


holger1411:
I´ve add "gulp-sourcemaps" to package.json. Thx!

PR #60, opened 2016-09-02
Revise markup for navbar toggle.

tedgeving:
1. Current BS4 documentation uses an entity for the navbar toggle instead of styled span tags.

2. Change navbar toggle class to match documentation. 

http://v4-alpha.getbootstrap.com/components/navbar/

### DIFF
--- a/header.php
+++ b/header.php
@@ -39,12 +39,9 @@
 
                                 <!-- .navbar-toggle is used as the toggle for collapsed navbar content -->
 
-                                  <button class="navbar-toggle hidden-sm-up" type="button" data-toggle="collapse" data-target=".exCollapsingNavbar">
-                                    <span class="sr-only">Toggle navigation</span>
-                                    <span class="icon-bar"></span>
-                                    <span class="icon-bar"></span>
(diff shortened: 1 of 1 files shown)


maxdmyers:
Just curious, why was this closed & deleted? This is the method used on the bootstrap docs and works well.

tedgeving:
Good question @maxdmyers,
I closed it because I was not sure about the correct process for submitting pull requests to the project.  

It looks like the existing navbar is using BS3 markup, the header should be updated.
https://github.com/holger1411/understrap/blob/master/header.php#L42

I'm happy to reopen it, let me know.

maxdmyers:
Please do, I found this PR to be valid and useful. If the theme is using B4, it should be fully embracing B4 markup.

tedgeving:
Please update the navbar to use BS4 markup.

holger1411:
I also add the hidden-sm-up class to reproduce the BS3 behavior. The hamburger icon will appear/work on smaller screens only.

holger1411:
+1 @tedgeving

PR #65, opened 2016-09-10
Add npm scripts

kamataryo:
Hello.
I like npm scripts.
This PR enables developers to set up with less commands.
How about this way?


holger1411:

PR #75, opened 2016-09-22
Fix gallery shortcode

kavillock:


flawiddsouza:
This works better what exists now. The images included in the gallery shortcode don't go missing if I used this. Can someone merge this? One suggestion: Removing `img-thumbnail` from `str_replace('<a', '<a class="thumbnail img-thumbnail"', $html);` would be good. Otherwise a weird rectangle appears behind the image.

Thomas-A-Reinert:
Is there a description of the problem and it´s solution available? Would love to help and put my "vote" in if neccessary.. ;)

flawiddsouza:
Using the current implementation, some of the images included in the gallery shortcode snippet just go missing and are not included for some reason. Though, I'm unaware of the real intention for this pull request. I'm using it in production and all's good but it might not be what everyone wants.

holger1411:
The initial pull request replaces the old gallery code with the code from Roots Sage Gallery (roots.io).
I have a bad feeling with just copy&paste the code from another WP framework.

But I´am more a Frontend and UX guy, not a PHP developer...so I can´t fix it or create a brand new gallery function from scratch by myself.

So any help is appriciated!

kavillock:
@holger1411 yes, I copy&paste this code from roots.io because old code not work properly (when we use 2 galleries in one post, gallery was duplicated with all images),

but in sage this file was deleted for sass equivalent:

see:
https://github.com/roots/sage/pull/1421
https://github.com/roots/sage/pull/1546

holger1411:
@kavillock Yes ... and I understand your intention. It solves the problem and its basically not a problem to use code from one open source framework in antoher. 
And there is no need to develope the same again and again...

I will try both solutions (find and fix the bug(s) or using it from roots.io) and will check it.
Just doing it like the roots.io guys (Using WP basic markup and create a CSS solution) isnt an option in my opinion. 
It means we need to create and maintain our own stylings for a gallery and a grid...while BS4 already provides anything we need for it...

Thomas-A-Reinert:
Sidenote: This perfectly works without touching any of the S/CSS! Your solution just doesn´t attach col-xx-X styles for some reason. And that seems to be all that´s missing.

holger1411:
Okay, going to merge this. Thx @all!

holger1411:
Hm, seems I need to revert this.
Wordpress.org did not accept themes with add_shortcode:

> WARNING: The theme uses the add_shortcode() function. Custom post-content shortcodes are plugin-territory functionality.

Sorry @kavillock !

stef-k:
I think this could be implemented programmatically so it should not be a problem, needs further investigation though.

PR #86, opened 2016-10-11
Fix version number in the footer

maxdmyers:
It would probably be better to dynamically inject the theme name and version into the footer, no? This method will work with parent or child themes.


holger1411:
+1 Thx to @maxdmyers

Issue #103, opened 2016-10-31
Official UnderStrap Feature Request List 


holger1411:
According tto @stef-k suggestion I open up an "Feature Request" list here. Comment any idea you have below this post. Add your +1 if you like a specific comment/request. If it sounds good and useful I will add it to this list. 
Features should be useful for the majority of UnderStrap developers/users. So too specific feature requests (for example for a very uncommon plugin) will not be added. Sorry!

# Feature Requests / Bug Hunt Backlog:

## 0.5.0 release:
- [x] Page template with sidebar on the left (Discussions here: https://github.com/holger1411/understrap/issues/116)

- [x] Clean (numeric) page pagination (Discussions here: https://github.com/holger1411/understrap/issues/127)
- [x] More languages
- [x] Make functions pluggable
- [x] WooCommerce templates for product pages, category, checkout etc.
- [x] Fixing translation in author template (Discussions here: https://github.com/holger1411/understrap/issues/107)

## Ongoing task(s):
- [ ] Creating better/more docs - maybe with GitHub pages(?)  :tada: (Discussions here: 

## Backlog (future releases):
- [x] Fixing the gallery or create a new gallery script  (Discussions here: https://github.com/holger1411/understrap/issues/115)
- [ ] Page template with "hero" owl.carousel area
- [ ] Full RTL support 
- [ ] One / some child theme´s with a unique design (Discussions here: https://github.com/holger1411/understrap/issues/113)
- [x] Implement/check coding standards and HTML/SASS syntax according to WordPress standards 
Discussions here: https://github.com/holger1411/understrap/issues/118)

stef-k:
I think it would be a good case to investigate and implement a flexible template that can be customized through the customizer api so that enables the user to choose left or right sidebar positioning.

holger1411:
The problem with a customizer setting is that you have sidewide just left sidebars or right sidebars. May make sense for LTR or RTL sites as a general setting, but you can´t choose/change it based on a specific page... and what if you need a left AND right sidebar just in some cases?

holger1411:
Maybe a general setting for the base template makes sense, but also adding custom templates for sepcial cases...will check some other WP themes and WP frameworks how they handle it.

maxdmyers:
I've submitted PR #106 to resolve the feature request "Implement/check coding standards and HTML/SASS syntax according to @mdo ´s http://codeguide.co/" above.

holger1411:
[...]

PR #135, opened 2016-11-17
default or masonry posts index

stef-k:
In a follow up of my [proposal ](https://github.com/holger1411/understrap/issues/103#issuecomment-259221513) this PR adds an option in customizer to choose between the default display style of posts index page (latest posts) and a masonry like style using bootstrap's card-columns and cards classes.

Ready for merge if you like it and agree for inclusion.

Additonally while implementing this, I've introduced some new files and features, here is the changelog:

* new file `home.php` which is responsible to show latest posts [see template hierarchy](https://wphierarchy.com/)
* new template in `loop-templates` the `content-card.php` that renders a single card
* a new file containing utility functions in `inc/utilities.php` currently has 1 function for custom length excerpts which it is used on cards. Additionally this could be used in Customizer as global setting if we implement it
* changed `functions.php` to load `inc/utilities.php`
* changed `inc/customizer.php` to include the new setting

screenshots of the two displaying modes (click on images for bigger ones):

**The default**
selection_001

**The new addition (masonry)**
selection_002

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -66,3 +66,8 @@ require get_template_directory() . '/inc/bootstrap-wp-gallery.php';
 * Load WooCommerce functions.
 */
 require get_template_directory() . '/inc/woocommerce.php';
+
+/**
+ * Load a utilities library.
+ */
+require get_template_directory() . '/inc/utilities.php';
--- a/home.php
+++ b/home.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * The template for displaying all pages.
+ *
+ * This is the template that displays all pages by default.
+ * Please note that this is the WordPress construct of pages
+ * and that other 'pages' on your WordPress site will use a
+ * different template.
+ *
+ * @package understrap
+ */
+
+get_header();
+?>
+
+<?php
+$container   = get_theme_mod( 'understrap_container_type' );
+$sidebar_pos = get_theme_mod( 'understrap_sidebar_position' );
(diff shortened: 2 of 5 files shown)


Thomas-A-Reinert:
First, let me thank you for your advice - in another post though. Will try to work through that somewhen later..

Yeap, I like the masonry idea and already got that integrated on my fork of the 3.x version of understrap as a page template, which also makes sense for portfolio like things, I guess. In that case it was just images with no text, so it looked pretty alright I guess.
But - and that´s just a personal opinion - once it includes text, it looks awfully cluttered and ain´t easy on the eyes, don´t you think so? What would you think of a flexbox solution with equal heights?

stef-k:
@Thomas-A-Reinert using flexbox to ajust heights from masonry becomes grid which is a different thing. IMO it should be a different option default-masonry-grid.

Additionally I would like one more option but for the momment i can't think how to implent something like this, any ideas would be welcomed. A vertical timeline display!

selection_001

Thomas-A-Reinert:
Sorry, @stef-k , you´re absolutely right. Not a native english speaker, as i said before, so please.. ;)
Anyway, masonry layouts with text still do look shitty from my point of view. A possible solution for that would be to display just images and a headline on mouseover which should be accomplished easily.

And I also do have something in mind for your timeline. Let me consider the possibilities for some time and I´ll probably come up with a solution, okay?

Sidenote: The timeline looks MUCH better with (longer) text involved compared to the masonry solution! But thats just my 2 cent..

PR #142, opened 2016-11-20
blog page as grid

stef-k:
As requested from @Thomas-A-Reinert here is a PR that adds the option to display latest posts (blog's main page) as grid.

Currently showing 3 posts per row without any option to modify it which should be easy to do from the customizer.

 I try to find if the customizer has the ability to conditionally show/hide secondary options. For example IF the Grid option is enabled, show the user secondary options which are only specific to the Grid display (posts per row, text- title colors, etc), until then I am leaving it as is.

Changes with this PR:

* `home.php` template changed to include the grid template
* `customizer.php` to add the new feature
* `content-grid.php` partial template that handles the rendering of individual posts

additionally changed how `get_template_part` is called.

Screenshot:

selection_001

### DIFF
--- a/home.php
+++ b/home.php
@@ -27,9 +27,8 @@ $posts_style = get_theme_mod( 'understrap_posts_index_style' );
 
 			<!-- Do the left sidebar check -->
 			<?php get_template_part( 'global-templates/left-sidebar-check', 'none' ); ?>
-			<?php if ( 'masonry' === $posts_style ): ?>
+			<?php if ( 'masonry' === $posts_style ) : ?>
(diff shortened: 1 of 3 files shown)


stef-k:
Update, the customizer now will automatically show or hide an additional control for selecting the correct column.

What it does: grid posts single post has column settings `col-xs-12 col-md-some_variable` this PR changes that variable.

Control hidden (grid posts are disabled)

selection_003

Control active (grid posts enabled)

selection_002

another one with different column setting
selection_001

PR #160, opened 2016-11-25
gulp dist: skip sass directory, gulpfile.js and package.json

saknopper:
When you consider that everything in the `dist` directory is a final product, those files/directories aren't needed. But feel free to correct me :)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -220,6 +220,6 @@ gulp.task('copy-assets', function() {
 // gulp dist
 // Copies the files to the /dist folder for distributon
 gulp.task('dist', function() {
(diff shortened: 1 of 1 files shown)


holger1411:
The /dist folder should include the install pack files which is available at wordpress.org and from https://understrap.com

The gulpfile.js and package.json file including the /sass directory is neccessary for the install pack to develope the theme and to work with. 
UnderStrap is not a ready made design theme but a theme framework. So 90% or more will/have to change it.
So they are a vital part of this framework.

If you want to use it as starter theme you should use it with the Child Theme. That comes with its own dev files so you don´t overwrite this files if you update the parent theme.

saknopper:
Ok, then I'm probably not using it as intended.

holger1411:
Hi!
You don´t need to use it as intended, but if the files are included you can use it in more ways than without. ;-)

PR #185, opened 2016-12-13
Update package.json

michelmany:
Add tether.js on dependencies to avoid error on console which Tether is required for Bootstrap Tooltips.

screen shot 2016-12-13 at 15 42 54

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "merge2": "^1.0.2",
     "browser-sync": "^2.18.1",
     "owl.carousel": "^2.2.0",
+    "tether": "^1.4.0",
     "undescores-for-npm": "^1.0.0"
   },
   "devDependencies": {}


holger1411:
Hi!
normally you don´ need to add thether as own dependency. Its already a BS4 dependency and is installed automatically as soon as you run:
`npm install`
if you run:
`gulp copy-assets`
it is copied to the /src folder and with
`gulp scripts `
it is added to the themes.min.js file

When/where do you receive this error?
Can´t reproduce it on demo site: https://understrap.com/understrap
and it was originally fixed somewhen in v 0.4.3

Make sure you run the three commands above to make sure you don´ overwrite new changes by an old dependency lineup.

michelmany:
Hi @holger1411 ,
I think I ended-up forgetting to run the `gulp scripts` then I was getting this error. I fixed it adding the tether direct to the package.json and running the three commands.

PR #201, opened 2016-12-29
Make Browsersync work correctly

VesterDe:
Removing reload entirely makes it possible for Browsersync to work correctly.

The error right now is that it's supposed to be injecting css into the site hot, without reloading the page for each css (.scss, .sass) file change. It does however reload on each change.

I've identified the issue to be the reload call at the end of the ['cssnano'] task. Removing it makes css inject into the site without the reload, while still making browsersync reload when other types of files change (like .php).

All other occurrences of reload seem to be obsolete as well, since it would only ever apply inside some form of 'watch' task in order to do any reloading. All other tasks with reload at the end are one-off tasks, without the need or possibility to do any reloading.

It's quite possible I've misinterpreted something though, so if anyone can check to make sure, that would be great.

P.S. the other commit fixes some typos.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -65,7 +65,7 @@ gulp.task('scss-for-prod', function() {
 
 // Run:
 // gulp sourcemaps + sass + reload(browserSync)
-// Prepare the child-theme.css for the developpment environment
+// Prepare the child-theme.css for the development environment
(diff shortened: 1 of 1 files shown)


holger1411:
Don´t see any problems with removing it.
Thx @VesterDe !

PR #218, opened 2017-01-17
Add BS typography styles to editor

ZXCVLuke:
I'd like to suggest some small modifications to the editor to better utilise Bootstrap. 

I have added a new php file for the functions. It includes the removal of h1 from the primary style dropdown because I believe it would incredibly rare for this to be used semantically while the post-title is already a h1. I have also enabled the style formats dropdown and added p.lead, small and blockquote styles so they may be used in the editor.

This is my first pull request and my php skills are pretty basic.

Thanks,
Luke

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -66,3 +66,8 @@ require get_template_directory() . '/inc/bootstrap-wp-navwalker.php';
  * Load WooCommerce functions.
  */
 require get_template_directory() . '/inc/woocommerce.php';
+
+/**
+ * Load Editor functions.
+ */
+require get_template_directory() . '/inc/editor.php';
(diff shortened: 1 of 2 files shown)


holger1411:
Good idea!
I will extend the editor markup selection and add more BS4 stuff in the future.
I just will remove the h1 removal. Its a good idea but there might be some usecases where an h1 is necessary (for example if you build a landingpage outside the normal page/article world...)

ZXCVLuke:
Thanks. Fair point on the H1 removal. I do it for my clients but it may not be for everyone.

ZXCVLuke:
And thanks for making UnderStrap. It's been great to use as a starter theme. I really like your handling of  all of the SCSS.

PR #221, opened 2017-01-19
changed it's to its

mikeritter:
use apostrophe for conjunction, not possession with its


holger1411:
Thx @mikeritter !
English is not my mother tongue. 
And to be honest: I choose the from of my "its" randomly...till now.

PR #283, opened 2017-04-14
Woocommerce 3.0.3 Support - Templates and Product Gallery*

typeplus:
Woocommerce 3.0.3 is a big change so there are many templates with changes or additions.

- Templates mostly straightforward
- There is still an issue of columns being 8.33% width especially on the "my orders" page but these are to do with stylesheets rather than templates.
- Product gallery 3.0.0 support added but the slider is not enabled as FlexSlider is required. I believe Woo recently made it so that you can use another slider (see here [https://github.com/woocommerce/woocommerce/pull/13009](url) ) but i dont have time to do this right now sorry.

Have tested these and appear to work. Been awhile since I used github so I am sure I buggered something up so apologies if proofing this is a chore.


typeplus:
Closing this as php formating issues need to be fixed on the master before this can be merged.

holger1411:
I merged your changes here:
https://github.com/holger1411/understrap/commit/fde19602236a8cf318568d121636deb2ebfd74fc
I fixed some problems...the build was failed because of some changes to the WP validator...even the underscores starter theme from Automattic failed the tests ;-)
So it isn´t a real problem at this point.

It will be in the final release 0.6.0 
Going to release this soon. There are just smaller issues I need to fix first.

Its already available if you use the big green "clone or download" button here on GitHub. Than you get the "bleeding edge" version.

PR #287, opened 2017-04-15
Woocommerce templates, Product Gallery & Phpcs fixes

typeplus:
Woocommerce 3.0.3 is a big change so there are many templates with changes or additions.

Templates mostly straightforward
There is still an issue of columns being 8.33% width especially on the "my orders" page but these are to do with stylesheets rather than templates.
Product gallery 3.0.0 support added but the slider is not enabled as FlexSlider is required. I believe Woo recently made it so that you can use another slider (see here https://github.com/woocommerce/woocommerce/pull/13009 ) but i dont have time to do this right now sorry.
Have tested these and appear to work. Been awhile since I used github so I am sure I buggered something up so apologies if proofing this is a chore.

Was getting TravisCI build fails so had a look at phpcs issues and hopefully they work.


holger1411:
Huge Thx @typeplus! I will merge this and do some testing. Will release it as soon as possible.

PR #291, opened 2017-04-21
my-address.php template

typeplus:
Fixes col-1 conflict (woo vs theme) which caused the 8.33% column width.


typeplus:
Something messed up here let me close this and fix.

holger1411:
Yes, I add my own Template for the address view.
The problem was Woocommerce uses a class called "col-1" and Bootstrap uses the same name.
I just removed this class because it was not necessarry.

typeplus:
Cool! Great work once again mate.

PR #298, opened 2017-04-26
Fix next/previous post links not floating correctly

kosvrouvas:
There is no **xs** in Bootstrap v4 for `.float-{sm,md,lg,xl}-{left,right,none}` so that makes the _nav_ links not floating.

See https://v4-alpha.getbootstrap.com/utilities/responsive-helpers/#responsive-floats


holger1411:
Ah, okay. It was removed in lapha5 because an xs class was obsolet.
Thx @kosvrouvas for fiddling this out!

PR #316, opened 2017-05-15
Bootstrap 4 float-xs no longer in use.

flyingchorse:
The class float-xs-left & float-xs-right no longer apply in Bootstrap 4. Had to create new function in my child theme in order to change classes


holger1411:
Hi!
Yes, since BS is flexbox only it did not work that way...I changed it here:
https://github.com/holger1411/understrap/commit/0cd0d3aca4109543a4c42f66fc9961d99a5322f2
with the BS´s flexbox utility class `.justify-content-between`

PR #327, opened 2017-05-22
Pagination Amend

raisonon:
Adds a class for next and prev buttons in the pagination so they can be styled differently if required.

### DIFF
--- a/inc/pagination.php
+++ b/inc/pagination.php
@@ -53,7 +53,7 @@ function understrap_pagination() {
 
 		/**    Previous Post Link */
 		if ( get_previous_posts_link() ) {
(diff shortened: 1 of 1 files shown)


raisonon:
Might make another amend related to the above and wanted to get feedback. Spotted there are also skip to end and skip to front buttons.

Screenshot: https://www.dropbox.com/s/a2eedsf7k228cha/Screenshot%202017-05-24%2013.30.42.png?dl=0

I've not seen these before in BS. Is this an Understrap addition? Think I might add classes to these too so we can hide with CSS. 

Any thoughts on this? Thanks

PR #346, opened 2017-06-08
Fix wrong escaping functions

ramiy:
When escaping HTML attributes we should use `esc_attr()`, not `esc_html()`.

This pull request replaces the wrong functions.


ramiy:
@holger1411 please review the files changed.

PR #367, opened 2017-07-13
Fix package name typo in package.json

zreese:
"underscores-for-npm" was missing an r


holger1411:
Sorry, no...there is a typo in the dependencies package.json
If you add the r you can´t download/update it anymore...

PR #418, opened 2017-09-24
Remove unwanted closed div tag in archive.php

ranukaherath:
archive.php had 4 starting div tags and 5 closing div tags .Removed additional one.


holger1411:
One div is opened in left-sidebar-ckeck.php. So there must be 5 closing tags

PR #459, opened 2017-11-17
Removed src folder from version control

kgoedecke:
In my opinion we should add the `src` folder to the `.gitignore`.
Those files are managed and dependent on npm and copied over during 
`npm install` or `yarn install`.

The README.md already contains instructions to run `npm install` so it should be good.

Looking forward for your feedback.

Cheers

Kevin


axlright:
I have the `src` directory removed from version control for my team. We each handle the dependencies on our own machines independently.

kgoedecke:
@axlright it makes sense I think! 👍

axlright:
Thanks @kgoedecke - I'm not that smart and kind of stumble through a lot of this stuff so I don't always know if my way of doing things is the right way.

I work out of a couple different co-working shops - Vuka / Impact Hub on the South side and WeWork downtown. Would love to say hi anyday. Coffee is free at both!

kgoedecke:
Let's see what @holger1411 's feedback on this is.

I just sent you an email, I usually work from Capital Factory. 2 min walk to WeWork haha!

holger1411:

PR #460, opened 2017-11-17
Moved npm dependencies to vendor folder and added to ignore

kgoedecke:
Hi Holger,

first of all, great work on the project! 

I was wondering if we can optimize one thing: I noticed that some dependencies which are copied during the `copy-assets` gulp tasks are still under version control. In my opinion it's good practice to keep dependencies out of version control.

I tried moving all files that understrap currently depend on to a separate `vendor` folder inside the general `/js` folder. 

See: https://github.com/kgoedecke/understrap/commit/49137ed3481ad01e514cfeb501e3b751d803068f#diff-b9e12334e9eafd8341a6107dd98510c9L251

This would have the huge benefit that a js dependency update doesn‘t effect the actual files and we can keep that out of the source control.

See my pull request below.

I found one problem, that you might have to help with or I have to open an other pull request:

The `functions.php` file in [underscores_for_npm](https://www.npmjs.com/package/underscores-for-npm) would have to adjust to the correct path as well.

Let me know what you think.


holger1411:
We tried this some time ago. The problem was that I was flooded by mails.
The problem is that many users did not use npm nor gulp via terminal.
They use GUI´s like Scout to compile the dist.
Without adding the src to the package they need to run npm install first.
It means I had to explain how to install and use npm constantly ...

So I re-added the /src folder again to the package and version control.
The basic idea behind this is that the understrap package should work out of the box, even if you did not use npm/gulp the "terminal way".

Don´t know if this is the correct way by-the-book, but saves me a lot of time ;-)

holger1411:

PR #471, opened 2017-11-27
Right Sidebar Check

Vishal-Deshpande:
Moved right sidebar check code to right-sidebar-check.php to follow DRY principle.


holger1411:
Sorry for my late reply.
Looks good and it is a usefull way to reduce code.
Need to test it first locally and will merge it as soons as possible.

Issue #496, opened 2017-12-11
Missing container class - row negative margin cause overflow-y


0dp:
Hi there,

Fresh install of understrap, using FF57. Don't know how or where this fit in, but `.row`s negative margin seems to create overflow on the y-axis.
<img width="1100" alt="screen shot 2017-12-11 at 22 41 49" src="https://user-images.githubusercontent.com/3642367/33855302-c8f91dca-dec4-11e7-919e-eab4a0d11d62.png">

Is this normal behavior and if so, how would you avoid it correctly?

craigpotter:
It's missing a container for me ......

0dp:
@holger1411 I think I got something here.
On a fresh install `get_theme_mod( 'understrap_container_type');` seems to return nothing. If I go to customize and select a container size, ` understrap_container_type ` gets set to `container`. If I try and change container type to "full width" `container` is not changed to `container-fluid`.

0dp:
in customizer.php:50
$choices = $setting->manager->get_control( $setting->id )->choices; 
Returns NULL.

0dp:
customizer.php:69
container_type should be `understrap_container_type`

antonapua:
same

antonapua:
#470 I shifted back to 6.5 until there are updates.

0dp:
@antonapua you should be able to apply the fix yourself. Check my comment or the PR i've made

holger1411:
Should be fixed now with 0.6.12 (parent theme) and 0.3.10 (child theme)

0dp:
I haven't tested yet, but it looks like it's the same

0dp:
Tested on v. 0.6.12 issue is still present

PR #498, opened 2017-12-12
container type never gets set

0dp:
#496


holger1411:
Closing this because of #500  was merged

PR #507, opened 2017-12-21
Update customizer.php

Lauren17:
Currently in understrap Version: 0.6.12 in the Theme Layout Settings the container type isn't setting. The above change sets it back to its default so it will set the whether or not the user wants a fixed container or full width container layout


0dp:
check PR #500 and #496

holger1411:
Closing because of #500 was merged - thx @0dp

PR #533, opened 2018-01-25
DRYed config and added a non-conflicting vendor option with cleanup.

lilolbear:
These are some streamlined changes that enabled me to build and manage several instances of under strap at once. Great work BTW.


lilolbear:
These last three commits were just adding in some ignores to make the travis ci build test a little cleaner. Since you want to include the vendor files in your src and js directories, I set it to ignore their errors for your build testing. Also added an ignore of the woocommerce directory to that list.

lilolbear:
These last three commits were just adding in some ignores to make the travis ci build test a little cleaner. Since you want to include the vendor files in your src and js directories, I set it to ignore their errors for your build testing. Also added an ignore of the woocommerce directory to that list.

PR #556, opened 2018-02-15
Add autoprefixer functionality

axlright:


IngoVals:
Shouldn't changes like this also be added to the child theme. I feel like the child theme process might be a bit inefficient because of all the dual code that is required. Every change to **Understrap** needs a evaluation about if it needs to go to the child theme as well, at least _gulpfile_ changes.

I can't pretend I have a better solution, since we are compiling the actual theme css from sass the only thing the child needs from the parent are the templates right? Perhaps we should be making the sass workflow, and hence the gulp commands in the parent since its easy to isolate the changes to the _sass/theme_ folder in a way that won't get overwritten on update and use the child theme for template and function.php changes. Would this be viable?

axlright:
@IngoVals an identical merge request was submitted for the child theme, and I assume that will be incorporated in a future update to the understrap_child theme.

To your point, I don't 100% understand the strategy of how the parent understrap theme and the understrap_child theme both contain all the vendor source Sass and JS. I think it's because @holger1411 and others on the team want people to be able to use the child theme out of the box without needing to use npm and gulp. But it does get into a situation where both the parent and child themes need updating. I need to understand the strategy more.

My team uses the understrap parent theme stock and we don't use / modify any part of it. We have this same autoprefixer running for the understrap_child theme where we do all Sass / JS editing and compiling.

I wouldn't want to edit the parent theme at all since it gets updated. But what could happen (and I think used to happen) is have the understrap_child theme reference the paren't theme's vendor Sass and JS.

I need to spend more time understanding it all before suggesting any changes. Hope this is helpful.

PR #572, opened 2018-03-01
Create he_IL.po

asaf147369:
hebrew PO file


asaf147369:
added po and mo files for the basic theme.
did all the translation with poedit and went over it to make sure everything is correct

Issue #568, opened 2018-02-28
gulp copy-assets don't work


fcojgodoy:
In a flesh installation I modify `gulpfile.js` like this:
```
gulp.task('scripts', function() {
    var scripts = [

        // Start - All BS4 stuff
        paths.dev + '/js/bootstrap4/bootstrap.js',

        // End - All BS4 stuff

        paths.dev + '/js/skip-link-focus-fix.js',

        paths.dev + '/js/parallax.min.js'
    ];
...
```
and, in `copy-assets` task:
```
// _s JS files into /src/js
    gulp.src(paths.node + 'parallax-js/dist/parallax.min.js')
        .pipe(gulp.dest(paths.dev + '/js'));

// _s JS files into /js
    gulp.src(paths.node + 'parallax-js/dist/parallax.min.js')
        .pipe(gulp.dest(paths.js + paths.vendor));
```

Then I run `gulp copy-assets` with no error but with no `parallax.min.js` in `./src/js` or `./js` folders.

What am I doing wrong?

Thanks to all the contributors!

P.D: Previously I install parallax.js package with `npm i -s parallax.js` and I check it is within `node_modules` folder.

fcojgodoy:
Hi!
The path is ok. The package name is parallax.js but the folder is parallax-js. I checked wth several packages.
If I remove the `var stream = ...` and `return stream` in the copy-assets task, the files are copied to the corresponding folders. @

0dp:
I can confirm that `copy-assets` does not copy all assets, and that removing  second `var stream` and first `return stream` "fixes" the task.

0dp:
@holger1411 @Thomas-A-Reinert please close

PR #589, opened 2018-03-05
Fix copy-assets

0dp:
Not all assets are copied when `gulp copy-assets` is run
#568

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -259,7 +259,7 @@ gulp.task('copy-assets', function() {
 
 // Deleting the files distributed by the copy-assets task
 gulp.task('clean-vendor-assets', function () {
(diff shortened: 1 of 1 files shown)


Thomas-A-Reinert:
Just asking for the sake of transparancy: Which one did not get copied/were missing?

axlright:
@Thomas-A-Reinert Looks like the change edited how the different font awesome font files were copied. I could be wrong. Usually am :-)

Thomas-A-Reinert:
@axlright all good in the hood ;)
Just asking to make sure that it´s not some "shady" stuff inserted but that everyone knows. This is really not implying that @0dp wanted to imply some "shady" stuff.

i just think - in general - that it would be good to know what the improvement is over the actually existing code.

And as you´re both belonging to the "member" team now I just thought we should follow a certain "code of conduct", if you know what I mean.

0dp:
@Thomas-A-Reinert @axlright The problem with the copy-assets task right now is that it has two return statements. It simply means that it copies the bootstrap stuff and exits. Instead of copying the bootstrap stuff, then the fontawesome stuff, then the _s stuff then exit.

I suggest that before merging, please make a fresh copy of UnderStrap and follow these reproduction steps.

1. delete content in the fonts folder.
2. run npm install
3. observe that nothing was copied over.
4. apply patch.
5. run npm install
6. verify that fonts where copied to the font folder

I will also post this in the original thread.

axlright:
@0dp @Thomas-A-Reinert can confirm that the patch works as intended based on the 1-6 steps outlined above. 

One note, the Embedded Open Type didn't copy - the original and the patched gulfile have .eot spelled .eof so that one font didn't copy over. I edited the file with that one character change and the Embedded Open Type did copy over. Was in the original gulpfile so maybe it was intended to be that way?

I don't know if I'm clever enough to submit a patch to a patch but I'll try!

Thanks for all your work.

PR #603, opened 2018-03-19
Added schema.org tags

rkb81:
I'd be adding the schema.org tags to header and content with a few extra customizations to title-tag and images adding the ```itemprop``` image to thumbnails.

### DIFF
--- a/css/custom-editor-overwrite-style.css
+++ b/css/custom-editor-overwrite-style.css
@@ -1,0 +1,3 @@
+body {
+    margin: 1vw;
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -186,7 +186,11 @@ gulp.task( 'scripts', function() {
 
         // End - All BS4 stuff
 
-        paths.dev + '/js/skip-link-focus-fix.js'
+        paths.dev + '/js/skip-link-focus-fix.js',
+
+        // Adding currently empty javascript file to add on for your own themes´ customizations
+        // Please add any customizations to this .js file only!
+        paths.dev + '/js/custom-javascript.js'
     ];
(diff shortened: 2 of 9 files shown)


axlright:
@rkb81 Thanks for this work! We use Schema markup pretty extensively especially for localBusiness and reviews to get those coveted stars in search engine results. 

It's my personal opinion that Schema markup is really ... personal and very specific to each project. Adding itemprops to items can generate as many errors as they do benefits depending on whether the items given itemprop are within a Structured Data type that supports that itemprop.

I do like the idea in the `understrap_schema_html_tag()` of more easily adding the itemscope itemtype into the <head> but it's my personal opinion that doing more may cause as many problems as it does solutions. I could certainly be the only one and I'm fine backing stuff out where needed.

Thomas-A-Reinert:
Hi @rkb81 , first things first: can you explain how my commits went into your pull requests?

Those are:
* css/custom-editor-overwrite-style.css
* gulpfile.js
* inc/editor.php

I guess you´ll understand that I won´t merge a pull request that includes files originally integrated by me, therefore I´ll decline it and ask you to issue a pull request again when removed. In general: please just only submit the files that are really needed and are in context of your PR.

Just checked your code and it seems pretty straightforward and shouldn´t pose a problem to whatever content being published. Microformats/Schema.org/Semantics are on top of my todo-list also. There were some things I recently deleted because they really *were* an issue and not performing as intended.

Anyway, I´ll review this a bit deeper in the near future when I find the time - but it already went on top of my list now. But please remove my stuff from your pull request ;)

Thanks in advance!

@axlright : would love to hear about your test results, too!

Issue #610, opened 2018-03-22
Pagination for get_terms


lordcameltoe:
Please delete anything that does not apply for you!

## I'm submitting a ...
  - [ ] bug report
  - [X ] feature request
  - [ ] suggestion

## Please tell us about your environment:

  - WordPress Version: 4.9.x
  - Browser: [ all  ]
  - Language: [ all ]

---

I've created a page that displays all the tags used on the site using the get_terms function.  I would like to add pagination to this page, however, I can't find documentation on how to use the understrap_pagination() function.

I have tried creating my own custom pagination using paginate_links(), however, I can't use the same styles use dwith understrap_pagination.

Is there any way understrap_pagination could work with get_terms ?

0dp:
To include the pagination into a custom page template use `<?php understrap_pagination(); ?>`

I'm working on a pagination that uses `paginate_links()` and incorporate BS4 styling maybe you can test it and modify it for your use case : https://github.com/0dp/understrap/blob/pagination/inc/pagination.php

What do you mean by used with get_terms?

0dp:
To answer the last part, this can work with pagination but you won't get numbered pages because max_num_pages return 0
Likewise, you'll have to remove the max_num_pages check in the current pagination.

```
$paged = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;

$args = array(
    //'taxonomy'               => 'category',
    'orderby'                => 'name',
    'order'                  => 'ASC',
    'hide_empty'             => false,
    'number'                 => 10,
    'offset'                 => $paged,

);

$the_query = new WP_Term_Query($args);

foreach($the_query->get_terms() as $term){
    echo $term->name." (".$term->count.")";
}
```
Correct me if I am wrong, but the short answer would be no. It's not possible with the standard WordPress pagination

lordcameltoe:
Here is my current code
```
$taxonomy = 'post_tag'; //get the tags
$number   = 3; // number of terms to display per page

// Setup:
$paged         = ( get_query_var( 'paged' ) ) ? get_query_var( 'paged' ) : 1;
$offset       = ( $paged > 0 ) ?  $number * ( $paged - 1 ) : 1;
$max   = wp_count_terms( $taxonomy, array( 'hide_empty' => TRUE ) ); 
$totalpages   = ceil( $max / $number );

// Setup the arguments
$args = array(
    'offset'       => $offset,
    'number'       => $number,
    'orderby'      => 'count',
[...]

PR #616, opened 2018-03-26
new pagination

holger1411:


0dp:
Would we consider #610 an edge case?

Thomas-A-Reinert:
Yup, I would, most definitely!

PR #619, opened 2018-03-28
Apply pluggable function wrappers

axlright:
Based on issue #550, applying pluggable function wrappers in one file to check with reviewers before continuing on to other files.

@holger1411 @0dp @Thomas-A-Reinert and anyone else take a look at this and let me know if you want it done differently. I don't know if you / others prefer `add_filter()` and `add_action()` to be above or below the functions. I am leaving as-is but I can make them all consistent. 

I also space things out a bit in php to make it more readable but I can cut down on the spaces if people prefer that.

PR #623, opened 2018-03-29
Update functions.php

0dp:


axlright:
I found a small typo in inc/pagination.php that was mended as a part of the #619 PR so maybe these should get reviewed / applied at the same time :-)

holger1411:
@axlright Yes, i merged your PR first.

PR #628, opened 2018-04-02
Add Polish translation

mirzal:
Subject says it all.

Translation is not complete, as I have skipped some strings that are no longer present in source. I guess `.pot` file could be re-created.


holger1411:
@mirzal Thx for your translation. Are you interested in joining the UnderStrap translation team and being responsible for the polish text?
If so let me know!

mirzal:
@holger1411 I think I should be able to handle this commitment. Can you tell me more how the process would work? Would you send me a message when there are new strings that need translating? Would I need to subscribe somewhere and keep track of changes myself?

PR #635, opened 2018-04-10
Fixed quite a few indentation and spacing issues.

gintsmurans:
Started to do this because page templates had strange indention for
#primary tag and thus was confusing - seemed like there was a missing
opening tag. Then saw few other indention issues and went through all
templates for the sake of consistency. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+          - ** Release 0.9.3 May 8th 2019 **
+          - hotfix release
+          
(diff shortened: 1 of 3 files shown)


malithmcr:
Hey @gintsmurans seems like there are some conflicts to fix.

gintsmurans:
This was a year ago, of course there are conflicts. :D 

Anyway, it looks like a lot of indentation issues are now fixed and there is no point of me going back to this.

Issue #632, opened 2018-04-06
Update both Parent theme and Child theme to use Bootstrap 4


dbybanez:
## Issue is about..
  - [ x ] a Question
  - [ x ] I need help

## Please tell us about your environment:

  - WordPress Version: 4.9.5
  - OS Version: Windows 10 
  - Browser: All
  - Language: PHP, MySQL, HTML, S/CSS, JavaSript, jQuery
  - Node.js & NPM Versions: Node.js v8.1.3 | NPM 5.1.0

---

Hi, 

I'm using UnderStrap v0.6.5 and the child theme v0.3.2. I wanted to update both the parent theme and child theme so I could use Bootstrap 4. Is there any documentation on how to migrate? Or any suggestions on how to migrate to Bootstrap 4?

Thanks,
David

axlright:
Hey @dbybanez in your case, since the change from child theme 0.3.2 to 0.5.1 is a big step, I'd recommend the following:

First update just the child theme. The parent theme contains all the php files but the styles and scripts for things like Bootstrap are all self contained in the child theme.

Copy the following files from the repo's child theme 0.5.1 into your child theme
gulpconfig.json
gulpfile.js
package.json

run `npm install` and make sure `gulp copy-assets` runs at the end of that process - it should automatically. Then run `gulp scripts` and `gulp styles` and you should be good to go. I'd then change the theme version number in your `style.css` to keep reference on where you are.

If / once that all works, I'd update the parent theme. You'll probably need to adjust some of your function names in your child theme `functions.php` and most likely the `new understrap_WP_Bootstrap_Navwalker(),` in your child theme `header.php` needs to updated.

Hope that helps.

holger1411:
And, if you use Bootstrap3 markup in your child theme .php files you have to adjust it.
Here is a migration doc:
https://getbootstrap.com/docs/4.0/migration/

pattonwebz:
Shameless plug here but the `understrap_WP_Bootstrap_Navwalker()` walker is based on a an old version of the [WP_Bootstrap_Navwalker](https://github.com/wp-bootstrap/wp-bootstrap-navwalker) that Ed, myself and a few other developers have maintained over the years. I have not tested the walker in this theme however the version which it was forked from did not have support for BS4 markup anyway.

The latest version of the walker at our repo supports bootstrap 4 (but does bring some breaking changes due to he corrected link-mod and icon class handling).

axlright:
[...]

PR #647, opened 2018-04-16
Swap customized walker to latest upstream wp-bootstrap-navwalker class

pattonwebz:
Use upstream navwalker in the theme. Ref #632


pattonwebz:
@Thomas-A-Reinert updated this PR. Should be working correctly now. I pushed it from my working directory this time like I should have done previously.

Note: `Understrap` is prefixed on the class name with an uppercase `U` You can change it to lowercase if you want so it matches the previous usage exactly but class names are not case sensitive so should work both ways :) This PR also adds 2 additional strings (with the text domain of `understrap`) in the fallback function that could need translated.

axlright:
Thanks @pattonwebz for your contribution 🚀

axlright:
@Thomas-A-Reinert I'm happy to help test in a demo environment if you can give me an outline of what to test. I'm kinda dumb about nav walkers, but I'd like to learn.

0dp:
> Thanks @pattonwebz for your contribution 🚀

EMOJIIIIIIII

Thomas-A-Reinert:
Works like a charm now and also keeps child-themes working after update!

Lesson learned: 

> PHP-functions are not case-sensitive

 ;)

Thanks again @pattonwebz !

Issue #631, opened 2018-04-05
Css & Js browser caching


gintsmurans:
## Issue is about..
  - [x ] a Feature Request

Hi! How about a proposal to add some kind of minified css and js file versioning. While developing a new theme I am getting tired of browser not always loading latest version, so as a quick fix I have modiefied enqueue.php like this:

    <?php
    /**
     * Understrap enqueue scripts
     *
     * @package understrap
     */
    
    if ( ! function_exists( 'understrap_scripts' ) ) {
    	/**
    	 * Load theme's JavaScript sources.
    	 */
    	function understrap_scripts() {
    		// Get the theme data.
    		$the_theme = wp_get_theme();
    		$theme_version = $the_theme->get( 'Version' );
    		
    		$css_version = $theme_version . '.' . filemtime(get_template_directory() . '/css/theme.min.css');
    		wp_enqueue_style( 'understrap-styles', get_stylesheet_directory_uri() . '/css/theme.min.css', array(), $css_version );

    		wp_enqueue_script( 'jquery');
    		wp_enqueue_script( 'popper-scripts', get_template_directory_uri() . '/js/popper.min.js', array(), $theme_version, true);
    		
    		$js_version = $theme_version . '.' . filemtime(get_template_directory() . '/js/theme.min.js');
    		wp_enqueue_script( 'understrap-scripts', get_template_directory_uri() . '/js/theme.min.js', array(), $js_version, true );

    		if ( is_singular() && comments_open() && get_option( 'thread_comments' ) ) {
    			wp_enqueue_script( 'comment-reply' );
    		}
    	}
    } // endif function_exists( 'understrap_scripts' ).
    
    add_action( 'wp_enqueue_scripts', 'understrap_scripts' );

holger1411:
@gintsmurans Thx! Will try and test it.
There are several issues about this

holger1411:
Add it here:
https://github.com/understrap/understrap/commit/4fb9fdd96dc34dc3d64356236b495e7da3de2a78
and it is live on our demo site:
https://understrap.com/understrap
If there are no problems/issues with your solution we will ship it with 0.8.2

Thx @gintsmurans

axlright:
[...]

PR #662, opened 2018-04-29
gulp rev

0dp:
We've been talking about asset versioning and cache busting over at #631 
This PR includes gulp-rev in order to create versioned assets without using query strings which gives negative scores on GT-METRIX et al. but also makes cache busting a breeze.

Please test before merge.


ZacharyElkins:
Does this export both a theme.min.css and a theme-[number].min.js ?

I ask because using this with a minify/combine plugin on manual (like W3 Total Cache) would mean having to type in the number change on each revision, unless both files are provided which it appears to do so.

0dp:
Hey @ZacharyElkins the new task does not interfere with the existing tasks ie. `minifycss` and `scripts`. Instead it is being executed after the two tasks when running `watch-bs`.

0dp:
This is PR is a bit sloppy. I'm going to close it and make a new PR tonight

PR #665, opened 2018-05-01
Fix to 'watch-scss'. Also adding some comments to other tasks.

pferlito:
'watch-scss' references non-existing 'scss-for-dev' task, which should be 'sass'.
Comments were missing on some tasks.


Thomas-A-Reinert:
@UnderstrapFramework and you close this because... ?!

Thomas-A-Reinert:
In general, any REASON why you close / merge a commit would be really helpful and more than just "common". Closing and/or merging a commit without any statement just sucks.
Personal opinion. Which does conform to the official standard though.

PR #666, opened 2018-05-02
gulp-rev

0dp:
Hi Y'all,

Made a new PR because the old one was a sloppy mess.

This is in response to #631

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -20,6 +20,7 @@ var cleanCSS = require( 'gulp-clean-css' );
 var gulpSequence = require( 'gulp-sequence' );
 var replace = require( 'gulp-replace' );
 var autoprefixer = require( 'gulp-autoprefixer' );
+var rev = require('gulp-rev');
 
 // Configuration file to keep your code DRY
 var cfg = require( './gulpconfig.json' );
@@ -94,6 +95,16 @@ gulp.task( 'cssnano', function() {
     .pipe( gulp.dest( paths.css ) );
 });
 
+gulp.task( 'rev', function() {
+  // by default, gulp would pick `assets/css` as the base,
+  // so we need to set it explicitly:
+  gulp.src([paths.css + '/theme.min.css', paths.js + '/theme.min.js'], {base: './'})
+    .pipe(rev())
+    .pipe(gulp.dest('./'))  // write rev'd assets to build dir
+    .pipe(rev.manifest())
+    .pipe(gulp.dest('./'));  // write manifest to build dir
+});
+
 gulp.task( 'minifycss', function() {
   return gulp.src( paths.css + '/theme.css' )
   .pipe( sourcemaps.init( { loadMaps: true } ) )
@@ -116,7 +127,7 @@ gulp.task( 'cleancss', function() {
 });
 
 gulp.task( 'styles', function( callback ) {
-    gulpSequence( 'sass', 'minifycss' )( callback );
+    gulpSequence( 'sass', 'minifycss', 'rev' )( callback );
 } );
 
 // Run:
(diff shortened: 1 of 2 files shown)


0dp:
If anyone has any good ideas on how we can manage revisions feel free to participate

MK-PL:
'gulp-rev' dependency is missing in package.json file ;-)

0dp:
Thanks @MK-PL will make a patch asap

Thomas-A-Reinert:
Hi @0dp - did some testing today and here´s what I can throw in..

## Issue is about..
  - [x ] a Bug Report

  - **WordPress Version:** 4.9.6-RC2-43289
  - **OS Version:** Windows 10
  - **Browser:** Chrome
  - **Node.js & NPM Versions:** Node.js [V 8.9.2] | NPM [V 6.01]


---

## Steps to reproduce?

Open up `./src/js/custom-javascript.js`.
Add the following lines for testing:
```javascript
window.onload = function(){ alert("Works!"));}
``` 

Everyone trying to use the new versioning system. At least using the new `custom-javascript.js`.

Versioning of the `custom-javascript.js`.

## *When* does this occur?
When adding / deleting new content to the custom-javascript.js the new versioned files won´t get regenerated.

See above.

## Are there any related Issues and/or PRs?
Nope

## Expected behavior
New version should a.) be generated b.) embedded.

Thomas-A-Reinert:
Yeap, I know. As I merged it believe, I did ;)

Thomas-A-Reinert:
Still:
```html<script type='text/javascript' src='//localhost:3000/test/wp-content/themes/understrap/js/theme.min.js'></script>```

When I change the content of the custom-javascript.js there is no new version being generated.

But I noticed another side-effect in Visual Studio Code. BrowserSync repeatedly runs without any code-changes. I guess that´s not intentionally?

0dp:
No that's bad news, I don't know if that is because I added `rev` to the watch-bs task.

But I can see the error. I accidentally wrote child-theme.min.js instead of theme.min.js

Gonna fix the patch in a minute

0dp:
I've corrected the patch, if you could test it plz.

Thomas-A-Reinert:
It doesn´t. It´s just an editor. Sorry if I wasn´t clear about that or caused any confusion. It has a built-in terminal, but that´s it. Just like phpstorm has.

And sorry again - I won´t find the time to check the update within the next few days.. :/

0dp:
Ok, it was just because you and another bloke were talking about some gulp setup in VS. Maybe that got me confused.

No stress, no worries. 🙏

0dp:
hi @pcarter13 
When gulp-rev runs, it creates a versioned version of theme.min.css and theme.min.js and stores the value in rev-manifest.json found in the theme root.
The files are then loaded in inc/enqueue.php 
Line 12 - 23 is a function that checks if rev-manifest.json exist and if it doesn't it returns an empty array. If the file exist it will looks at the key and parse the value when the scripts are enqueued.
That means, if anything goes wrong, asset_path will just return theme.min.css or theme.min.js.

> my colleagues view the site they still have to use a hard refresh to see the changes
When they visit the site, what file is being loaded before they do a hard refresh? It could be a caching thing (ie. squid) - Could you maybe ask a colleague to visit the site with the network tab open in the browser and record the network transactions? (please post screenshot)
The reason I say this is - If a new file is created and the link to that file is replaced. Then there is no reference in the code to the old file. Hence, if the old file is loaded it must be a network thing...I think

pcarter13:
Am I supposed to be able to see the custom name of the CSS file? Because even on my machine the network tab is shown as such, just regular theme.min.css

screen shot 2018-06-25 at 14 46 09

0dp:
Yeah, you're supposed to see the rev'ed version in your source.

Could you do me a favor and and change line 14 in inc/enqueue.php to
`if ( !empty($manifest_path ) ) {`

PR #667, opened 2018-05-03
BS4 Individual JS Modules

ZacharyElkins:
This allows for the individual BS4 JS files to be used.

- Leaves the pre-compiled BS4 JS file as the default.
- To use individual files comment out the pre-compiled BS4 file and remove comments before the modules required for your project.

This is using the `js/dist` files from Bootstrap. These individual modules are ready to go, only requiring `util.js`, `jquery.js`, and some require `popper.js`.

The alternative way to do this is to use something like Babel and Rollup to compile the `js/src` files together. The alternative method would mean a smaller export file, as it adds the "glue" to the mix and removes unnecessary code from the individual files. I have a separate branch which uses the alternative method but **would like some feedback on which one to include with Understrap**.

The Babel/Rollup way is more intensive and will add several dependencies and an additional gulp task.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ node_modules
 
 understrap.zip
 
-dist
+/dist
 
 dist-product
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -139,7 +139,25 @@ gulp.task( 'scripts', function() {
     var scripts = [
 
         // Start - All BS4 stuff
+
+        // Full BS4 JS file. Comment this out if you are using the individual modules below
         paths.dev + '/js/bootstrap4/bootstrap.js',
+
+        // BS4 Modules. Select which BS4 modules to include.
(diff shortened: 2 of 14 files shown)


danemorgan:
We seem to have gone the Babel route on this.

0dp:
Well i do agree with @ZacharyElkins on this improvement.
But there is no need to pull the individual modules into src via copy-assets.

Instead the files can merely be listed in the scripts task, so that you can comment out BS modules you don't want to include. 

Or you can have a file in src with all the imports from BS and you can uncomment there. Then only include one extra file in scripts task.

danemorgan:
@0dp Should we reopen and discuss further. I'll admit freely that this is in some uncharted waters for me. I may have assumed something that wasn't justified here when I said it seemed that we had chosen a different course.

0dp:
No worries.
I can make a PR and you can see what i mean

Issue #663, opened 2018-04-29
Pagination throws error


ZacharyElkins:
Please delete anything that does not apply for you!

## Issue is about..
  - [ ] a Bug Report

## Please tell us about your environment:

  - **WordPress Version:** 4.9.x
  - **OS Version:** [ Windows 10] 
  - **Browser:** [ all ]
  - **Language:** [ PHP ]

---

# Issue Overview for Bugreports
Pagination gives arror of "Invalid argument supplied for foreach() in LINE 40"

## Steps to reproduce?
Use pagination. I am using WooCommerce, if that has any effect.

## *Who* is the bug affecting?
All users.

## *What* is affected by this bug?
Pagination doesn't show, instead throws an error.

## *When* does this occur?
Ever time pagination attempts to load on page.

## *Where* on the platform does it happen?
Anywhere pagination occurs. Search, listing products, listing blog, etc.

## *How* do we replicate the issue?
Use pagination, check. If not seeing an issue, use WooCommerce.

## Are there any related Issues and/or PRs?
Pagination was changed here @0dp :
https://github.com/understrap/understrap/commit/33183a34811f4f02fd235490524b71d844bb278a#diff-80ab22429bece23e40711f16bb5f3a64

## Screenshots / Video
![pagination](https://user-images.githubusercontent.com/25933268/39410026-26abb10c-4bae-11e8-8979-ceb16adca4f8.JPG)

## Expected behavior
Pagimaton should generate.

# Todos
Fix PHP

 * Test
 * Documentation

### Other Comments
I beleive this is because of the use of wp_parse_args. When this is removed, or a seperate array of args is provided without it, it works as expected. Attempts to fix the error specifically by rewriting the foreach statement end in rendering without error but improper display/output.

0dp:
So I don't know if WooCommerce has any effect on the pagination, but you said that WooCommerce uses its own breadcrumbs. Maybe it's the same for pagination?

What happens if you disable WooCommerce?

I have the pagination working just fine in my end on my UnderStrap development site with WooCommerce activated.
<img width="757" alt="screen shot 2018-04-29 at 21 55 05" src="https://user-images.githubusercontent.com/3642367/39410429-643a967e-4bf8-11e8-9ffe-0bb7b8cf2d7a.png">

On line 40 the foreach loop, loops through $links from line 23.
Could you try and see what the contents of $links is on one of those archives?

[...]

Issue #552, opened 2018-02-12
Some issues I found


DumboJetEngine:
Hello.
While trying to use this theme, I came across some issues:

1. It seems that there are two versions of the file `pagination.php`. I am not sure why this happens, but it's weird and it looks like leftovers. These are the files:
understrap\inc\pagination.php
understrap\global-templates\pagination.php

---

2. While using the `understap-child` theme, I had to modify the file `understrap-child/sass/assets/understrap.scss` from this:
```
@import "../../src/understrap/understrap/understrap";
```
...into this (so that the SASS can be compiled):
```
@import "../../sass/understrap/understrap/understrap";
```
I am using this file as a compilation root:
```
understrap-child/sass/child-theme.scss
```

There were some weird problems with `contact-form7.scss` not compiling, but that looked weird since the SASS code seemed fine to me. Might be an outdated SASS compiler or something... Anyway, I have removed that file for now....

---
3. The pagination UI is not very sound or at least it looks weird (I am currently working on a fix):

![image](https://user-images.githubusercontent.com/8676903/36122287-7ab4c524-1051-11e8-96b0-47fe98d1c47f.png)

holger1411:
Hi @DumboJetEngine !
Would be great if you could spend some time on the pagination issue.
Seems that both pagination.php files where added/edited with commits and that that caused some trouble.

To 2:
The correct @import path should be:
`@import "../src/sass/understrap/understrap/understrap"; `
Like in this file: https://github.com/holger1411/understrap-child/blob/master/sass/child-theme.scss#L3

0dp:
I'm looking into adapting `get_the_posts_pagination();`

0dp:
I've created a branch in my fork (not sure how I can create a branch here) 
https://github.com/0dp/understrap/tree/pagination
That implements what I believe is a more current pagination. Please note I still consider this a P.O.C.
There is definitely a few things that can be improved upon like custom arrows

holger1411:
@0dp : I opened a pull request and will test it:
https://github.com/understrap/understrap/pull/616

BenAttenborough:
I've noticed that pagination with dots is broken, in that it doesn't take into account the ellipsis when working out the page number on the pagination:

![screen shot 2018-05-09 at 16 28 15](https://user-images.githubusercontent.com/8101421/39823941-0804318e-53a6-11e8-9852-0922381737ad.png)
(understrap_pagination() above, paginate_links() below)

[...]

PR #668, opened 2018-05-05
dynamic prev_next

0dp:
I've noticed that with the current implementation, next arrow is still shown if there are no more pages.
By setting prev_next to true, WordPress should handle the generation of next and previous arrows. 
This also means that the code is dramatically simplified.

But please test this, because I don't remember why I didn't write the code this way in the first place.
Maybe because there was some fringe case I tried to rectify?

I'm just referencing this issue #663 just because it made me look at `understrap_pagination` some more

### DIFF
--- a/inc/pagination.php
+++ b/inc/pagination.php
@@ -12,7 +12,7 @@ if ( ! function_exists ( 'understrap_pagination' ) ) {
 
 	    $args = wp_parse_args( $args, [
 	        'mid_size'           => 2,
-	        'prev_next'          => false,
+	        'prev_next'          => true,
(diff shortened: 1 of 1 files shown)


0dp:
We'll just wait with this PR until we get response on #552

0dp:
@holger1411 @Thomas-A-Reinert @holger1411 Please merge this, since it addresses serious issues with the current pagination.

PS. The travis check has failed because it cannot connect to MySQL?

Thomas-A-Reinert:
Re-ran the test and it passed, @0dp . Merging it now.

PR #677, opened 2018-05-17
Proposal fix for missing js rev

0dp:
@Thomas-A-Reinert 
#666  😈


0dp:
@Thomas-A-Reinert try and see if that works better for the revision of of theme.min.js
If the continues running of `watch-bs` keeps happening. Try and remove `rev` from the `watch-bs` task.

PR #678, opened 2018-05-18
Added cookies checkbox support for inc/custom-comments.php

JPKCom:
Since WordPress 4.9.6, there is a new checkbox in the comment form template needed:
https://wordpress.org/news/2018/05/wordpress-4-9-6-privacy-and-maintenance-release/

This pull request adds support for this field in inc/custom-comments.php

Maybe this field also needs to be added to the themes translation files.

### DIFF
--- a/inc/custom-comments.php
+++ b/inc/custom-comments.php
@@ -24,15 +24,17 @@ if ( ! function_exists( 'understrap_bootstrap_comment_form_fields' ) ) {
 		$aria_req  = ( $req ? " aria-required='true'" : '' );
 		$html5     = current_theme_supports( 'html5', 'comment-form' ) ? 1 : 0;
 		$fields    = array(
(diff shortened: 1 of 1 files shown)


0dp:
Really dig the GDPR efforts being put into WP atm.

Thomas-A-Reinert:
Nope, I don´t @0dp . That law is in effect for about TWO FRIGGING YEARS! And they meet the deadline like ONE WEEK BEFORE?! Congratulations, really! 👎

0dp:
GDPR is in effect from the 25th of may this year.
We have known for years that it would come. But lets all just agree that GDPR is a load of shit, and no one from the committee actually came out with any usable guidelines. Instead we've been trying to figure out what it actually means up until recently.
The fact that two guys decides to figure out what it means, and how it affects WordPress - Then implement it into WP using their company's lawyers, time and money. That is surely something appraise.

PR #689, opened 2018-05-26
Use template instead of stylesheet directory

AMDG2:
Hello !

I tried to use understrap as a parent theme for my latest wordpress work. I realize that a simple theme which is not using any builder to build a complete CSS stylesheet from the understrap one is broken.

Using the template directory instead of the stylesheet directory to include the theme stylesheet allows this simple child theme usage and doesn't break when only using understrap.

Let me know if this is ok for you.

Regards
Baudouin

### DIFF
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -28,7 +28,7 @@ if ( ! function_exists( 'understrap_scripts' ) ) {
 	 */
 	function understrap_scripts() {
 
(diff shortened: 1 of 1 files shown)


danemorgan:
I may be missing a step here, but as I read this, it would break enqueueing the parent styles from a child theme in the standard way because it would just look for the file in the child theme's URI again.

Combined with the staleness and the conflicts, I am closing this one.

PR #700, opened 2018-06-06
Woo commerce 3.4.0

ZacharyElkins:
This updates the WooCommerce templates for any changes that were made from 3.3.0 to 3.4.0

3.4.0 is the current main release.

## Notes
1. Any files not included in this PR were checked and had no updates - meaning those files will appear to have outdated version numbers, but they are the correct and newest version of that specific file.
2. I included links to the commits made by WooCommerce for each file, under reference.
3. I removed the add-to-cart button function, as including the file may be easier for users.
4. Commits are organized per single file, I figured this would be easiest for review.

### Edit:
It looks like Github didn't render the SHA as links under reference. I apologize, I thought I had it working. To fix this will mean a whole lot of work, so let me know if that is something everyone wants. The reference is there in case people want to review the changes by seeing the history of the Woo source files on their repo. I've gone through each one and verified these changes though.

@holger1411  @Thomas-A-Reinert

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -13,7 +13,7 @@ if ( ! function_exists( 'understrap_woocommerce_support' ) ) {
 	 */
 	function understrap_woocommerce_support() {
 		add_theme_support( 'woocommerce' );
-		
+
 		// Add New Woocommerce 3.0.0 Product Gallery support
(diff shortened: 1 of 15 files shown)


ZacharyElkins:
@holger1411 @Thomas-A-Reinert I'm not sure anyone is going to review these. I've looked over them a few times and am confident the changes are correct.

PR #701, opened 2018-06-06
WooCommerce Style Fix

ZacharyElkins:
This adds a new SCSS file for WooCommerce. This way if anyone is not using WooCommerce they can easily remove it from their theme.

It also incorporates a fix for the wrapper `woocommerce-input-wrapper` that was added to billing fields. It's either this or we remove the wrapper.

### Reference:
https://github.com/understrap/understrap/issues/698

### Notes:
1. `gulp styles` changed one item from `transparent` to `rgba(0, 0, 0, 0)`. I'm not sure why this occurred, but they are equivalent values. If anyone has information as to why, I would appreciate it.
The exact changes were:
`  -webkit-tap-highlight-color: transparent; } `
`  -webkit-tap-highlight-color: rgba(0, 0, 0, 0); } `
Seems like it has to do with auto-prefix.


0dp:
Interesting with the conversion from transparent to rgba. Will definitely investigate

Thomas-A-Reinert:
Guess why? If ANY browser has major problems with _any_ standards which is that? The browser from hell: SAFARI.

PR #702, opened 2018-06-06
- Minor grammar and spelling corrections.

davidshq:
- Updated copyright dates.
- Added introduction explanation of installation methods, was a little confusing at first glance whether all three of these needed to be performed or only one, thought this would help clarify.
- Added a similar section to Page Templates to help clarify purpose of section.


Thomas-A-Reinert:
Thanks for every submission - no matter as "small" as it can be!

Issue #708, opened 2018-06-10
Moving Sidebars Into a Sub-Folder?


davidshq:
Hi All,

Just wondering if it would make sense to move the sidebars into a sub-folder? There are six of them currently and seems to be cluttering up the main directory a bit. Just a thought!

Dave

Thomas-A-Reinert:
Yes, I guess none of us ever thought about that. But you´re absolutely right.  Doesn´t make sense to have them in the main dir. Thanks for suggesting that, really. Will try to fix that asap. (Can´t promise though, cos i ´m REALLY REALLY busy atm.) 

No promises being made, that one went instantly on my top priority list though ;)

axlright:
Added PR #711 to accomplish this. I have about 12 minutes of free time a day, so this seemed like a good use of those minutes!

PR #711, opened 2018-06-12
Organize sidebar files into loop-templates directory

axlright:
Based on issue #708, moved sidebar files into `loop-templates` directory. Deleted sidebar.php as no files actively referenced it (and it had old markup). Made two other minor closing div comment changes.


Thomas-A-Reinert:
Hi Alex, don´t want to be "über-correct", but the sidebars are not really LOOP-Templates, are they? So I think they´re kinda misplaced in the loop-templates dir.

I´d suggest putting them in the global-templates dir or into a new sidebar-templates dir, what do you think?

axlright:
I'm down to move them. In most of my client projects, I put everything except page templates in `loop-templates` because I see them as sub-actions of a page template. Whether there is technically an official `while` loop or not. Template parts, for each loops, traditional loops. It's all kind of the same to me. But that's personal / stylistic (which is what all organization is).

I don't personally understand what is meant by `global-templates` but that seems like the best destination if `loop-templates` isn't widely desired. I think making a third directory for `sidebar-templates` is one too many directories considering there are already sidebar functions / template parts in `global-templates`

ZacharyElkins:
I would vote for them being in `global-templates` as `loop-templates` does have a specific definition that they do not fall in.

I wouldn't mind another dir just for the sidebar templates either.

davidshq:
+1 to another folder just for sidebar templates. :)

axlright:
@ZacharyElkins and @Thomas-A-Reinert and @davidshq updated the PR to move sidebar template parts into a `sidebar-templates` directory. Thanks for the feedback!

Thomas-A-Reinert:
👍

PR #722, opened 2018-06-26
fix for syntax error

sage393:
Issue: "missing ) after argument list" on line 24


sage393:
You're welcome @0dp @Thomas-A-Reinert

PR #729, opened 2018-06-28
Duplicate revision task

0dp:
I've removed the duplicate revision task.

Also revision was only called when trying to make a new dist. This is wrong, it should run every time a css or js file changes.


0dp:
I just updated the patch, and tested. It works in my end

PR #745, opened 2018-07-16
Browser-Sync loop

0dp:
minor fix to prevent Browser-Sync loop

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -36,8 +36,8 @@ function scss( ) {
 exports.scss = scss;
 
 // Minify CSS
-function minifycss( done ) {
-    gulp.src( paths.css + '/theme.css' )
+function minifycss( ) {
+    return gulp.src( paths.css + '/theme.css' )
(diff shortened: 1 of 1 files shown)


0dp:
No one is gonna test this?

gulpfile.js
```diff
@@ -65,7 +63,9 @@ function scripts( done ) {
         paths.dev + '/js/skip-link-focus-fix.js',
         // Adding currently empty javascript file to add on for your own themes´ customizations
         // Please add any customizations to this .js file only!
-        paths.dev + '/js/custom-javascript.js'
+        paths.dev + '/js/custom-javascript.js',
+
+        paths.dev + '/js/search.js'
```
holger1411: Where comes this search.js from ?

0dp:
That reference to search.js is garbage I removed.
The reason why it was there is because I used this particular gulpfile.js in production, with positive results.

PR #749, opened 2018-07-19
How can i update the theme ? 

ukthemes:
I have used this theme . I did use main theme file and version 0.8.1 . I did change many code from many file . Have any easy way for update to latest version ? 

Advance Thanks


holger1411:
The best practice is to use a child theme for your changes and the parent theme for the global changes.
So you can separate both.
And pls open an issue instead of opening a pull request.
Pull request means you request that your code should be pulled into UnderStrap ;-)

PR #762, opened 2018-08-05
Adding .idea to .gitignore?

davidshq:
I know we don't want to add every exclusion anyone could possibly use to the .gitignore file for understrap, but might it make sense to add .idea as an exclusion? I'd think an exclusion for vscode might be worthwhile as well, as I imagine a significant portion of the understrap user base uses one or the other.


redpik:
:+1:

PR #777, opened 2018-08-20
Strapify select for WP default widgets

IanDelMar:
Eventually, simply using 
```scss
select { 
	@extend .form-control;
}
```
to capture all  `select` would be an option too.


IanDelMar:
Hi @holger1411! What about this one?

PR #782, opened 2018-08-22
Remove unnecessary call to count()

IanDelMar:
`count( $sidebars_widgets_count[ $sidebar_id ] )` called twice though returned value from first call stored in `$widget_count`


IanDelMar:
Closing this - see #796

Issue #796, opened 2018-09-04
Compatibility issue polylang


IanDelMar:
## Issue is about..
  - [ ] a Bug Report
  - [ ] a Feature Request
  - [x] a Suggestion
  - [x] a Question
  - [ ] I need help

## Please tell us about your environment:

  - **WordPress Version:** 4.9.8

### Issue
I'm currently working on a multilingual website using the free version of the plugin Polylang (2.3.10). In order dynamically handle css-classes according to the number of widgets in the footerfull and statichero sidebars UnderStrap uses `understrap_slbd_count_widgets()` in inc/widgets.php. Which works fine as long as the sidebar does not contain widgets of different languages.

Using Polylang you can choose whether a widget shall be displayed for all languages or a specific language. Polylang then displays the widget for the chosen language only via a frontend filter. This is where `understrap_slbd_count_widgets()` fails as it counts the unfiltered widgets and thus gives the incorrect count and css-classes. Example: You have 2 text widgets one saying "good morning"/"buon giorno" and one saying "good evening"/"buona sera", then `understrap_slbd_count_widgets()` gives you the classes for a 4-widget case while only 2 widgets are displayed at the same time.

### Potential solution
To overcome this one could call `understrap_slbd_count_widgets()` from a template file which would imply a rewrite of the concerning template files and maybe addition to the scss . Hence, it's not an option. The other way I found so far is to use the `dynamic_sidebar_params` hook like so

```php
/**
 * Add filter to the parameters passed to a widget's display callback.
 * The filter is evaluated on both the front end and back end!
 *
 * @link https://developer.wordpress.org/reference/hooks/dynamic_sidebar_params/
 */
add_filter( 'dynamic_sidebar_params', 'understrap_widget_classes' );

if ( ! function_exists( 'understrap_widget_classes' ) ) {
	/**
	 * Count number of visible widgets in a sidebar and add classes to widgets accordingly,
	 * so widgets can be displayed one, two, three or four per row.
	 *
	 * @since x.x.x
	 *
	 * @global array $sidebars_widgets
	 *
	 * @param array $params {
	 *     @type array $args  {
	 *         An array of widget display arguments.
	 *
	 *         @type string $name          Name of the sidebar the widget is assigned to.
	 *         @type string $id            ID of the sidebar the widget is assigned to.
	 *         @type string $description   The sidebar description.
	 *         @type string $class         CSS class applied to the sidebar container.
[...]

PR #802, opened 2018-09-10
Fix for #796

IanDelMar:


Thomas-A-Reinert:
Hi @IanDelMar  - sorry, never met such a problem with WPML or such.
I guess that need thorough testing by the core team. I personally do not use Polylang - anyone else of core team using it? Please don´t get frustrated if it takes a while - I guess we´re all currently working on projects, so your submission won´t get declined until further testing by anyone able to reproduce it.

By the way: thanks for helping out in various issues. This didn´t go unrespected ;)

IanDelMar:
@Thomas-A-Reinert thanks for your feedback and no worries I'm very well aware that we're all doing this in our free time. For now I just deregistered the sidebars and reregistered them inside my child theme, so there's no hurry anyway.

PR #804, opened 2018-09-12
Fix wrong positioning of add_action

IanDelMar:


pattonwebz:
LGTM

IanDelMar:
I think it's best practice to have the `add_action( ... )` outside the `if ( ! function_exists( ... ) )`.

For example if a child theme does
```php
function understrap_add_site_info(){
	echo 'Oops';
}
```
the action and all functions hooked to it won't fire.

PR #835, opened 2018-10-23
Templates directory

ylkyrg:
Create templates/ to declutter the root.


noelspringer:
I am also in favour of decluttering the root directory by using a single `templates` directory.

However, rather than having multiple directories in the `templates` directory I would like to suggest renaming the template files to indicate their use.

For example all the `sidebar-templates` files already have the prefix `sidebar-` (apart from `sidebar.php`) and all the `loop` templates already have the prefix `content-` (apart from `content.php`).

A similar approach could be used for the global templates by prefixing the templates with `global-` and the same could be done for `page-templates`.  Then all template files could reside at the root of the `templates` directory and the name of the file would clearly indicate its use.

Thomas-A-Reinert:
I totally understand the intention and endorse it personally.
Anyway, I vote against it, because it will most likely f**k up almost any installation that relies on child themes.

ylkyrg:
@Thomas-A-Reinert I understand, that's why I didn't implement what @noelspringer suggests, it's an easier fix on child themes. I could probably add backwards compatibility with get_template_part_{$slug} if anyone thinks it's worth it.

Thomas-A-Reinert:
Maybe I'm missing out on something but both suggestions will break existing child themes, right?

So don't get me wrong, but as long as there is no solution at hand that will keep existing child-themes (with current structure) working after an update, that's a showstopper for me personally. As there is already a comparably high amount of sites running on understrap, we cannot just introduce a "cleanup" from one day to the other that will break the majority of site.

Still - I'm fully backing your idea. Hope you also understand my concerns.

ylkyrg:
Sorry I was't clear. Yes, both suggestions will break child themes. The fix is slightly easier in my pull request than in @noelspringer's suggestion. That's why I did it this way.

Separately, we could probably add backwards compatibility using the get_template_part_{$slug} action hook if the gains are deemed worth it. I will update the pull request when I can.

Thomas-A-Reinert:
Cool, thanks for your efforts. Please just make sure that the new paths will also not kill existing childs. It's not just about the naming.. ;)

noelspringer:
All good. I hadn't thought about child themes. Cheers

ylkyrg:
I have moved all templates to templates/. I didn't give global templates a global- prefix as there is no global.php that it would make sense for them all to fall back to.

My attempt at backwards compatibility will look in the relevant directories (sidebar-templates/, loop-templates/, global-templates/) before including anything from templates/. The problem is I can't stop get_template_part() from executing, so when get_template_part( 'templates/sidebar', 'left' ) is called, if a child theme is active which has a sidebar-templates/sidebar-left.php, both are included. Maybe someone else can figure out how to make it work.

Otherwise, a custom understrap_get_template_part() could be written or the templates could be moved without backwards compatibility. I note the sidebar templates were previously moved without backwards compatibility so perhaps this wouldn't be the end of the world.

PR #851, opened 2018-11-15
formatting

0dp:


IanDelMar:
Hi @0dp! This has already been fixed in the current version of `inc/enqueue.php`.

PR #859, opened 2018-11-23
Fix dropdown when setting depth=0 in wp_nav_menu

stevygee:
depth=0 is the default and should display the entire hierarchy according to WP docs. Without this change, you'd have to explicitly set depth to 2 or above in order for submenus to be displayed. With this change, depth=0 also works as expected.


0dp:
Yoda condition I see. Change it we must

pattonwebz:
@stevygee if you are able a PR on the source repo at wp-bootstrap/wp-bootstrap-navwalker would be much appreciated :)

stevygee:
Oh I wasn't aware of the other repo. Just checked and there is already an open PR https://github.com/wp-bootstrap/wp-bootstrap-navwalker/pull/385

PR #868, opened 2018-12-11
Woocommerce Form Lost Password 3.5.2 update

Fatshape:
Added do_action( 'woocommerce_before_lost_password_form' ); and do_action( 'woocommerce_after_lost_password_form' ); 

Removed wc_print_notices();


Fatshape:
I made a diff with the original [Woocommerce template](https://github.com/woocommerce/woocommerce/blob/master/templates/myaccount/form-lost-password.php) and found they are no longer using it.

PR #883, opened 2018-12-28
New and improved gulp4 workflow

0dp:
All dev assets now live in ./src
That means there is no ./sass folder anymore. Instead it lives in ./src/scss/

There is currently no need for theme.css/js and theme.min.css/js
Instead files are being built according to task. So `npm run build` will create minified and uglyfied versions of theme.css/js

Understrap now support ES6
Components can be created in ./src/js/components and should be imported in ./src/js/theme.js
```
// Here you can import custom reusable components
import './components/component.js';
```
In ./src/js/theme.js you can also include regular js and jQuery scripts
```
// Here you can import custom scripts
import './custom-scripts.js';
```

You do not need to make workarounds to use $ instead of jQuery. This should be handled automatically by webpack. Yes there is webpack and gulp 😬 

### Available Commands

> Please make sure you are running the latest version of gulp-cli (2.0.1)
> Please verify with `gulp -v`

`npm run start` watch, compile and reload. replaces `gulp watch bs`
`npm run build` same as `gulp dist`
`gulp styles` compile scss to css with sourcemap
`gulp styles --prod` compile scss to css minify etc. without sourcemap
`gulp images` compress imagess and move to ./images
`gulp scripts` compile js, es6 components and minify
`gulp clean` delete dist
`gulp dist` build dist
`gulp watchForChanges` replaces gulp watch since watch is reserved
`gulp copyAssets` replaces copy-assets

PR #885, opened 2019-01-02
Remove .skip-link

IanDelMar:
The functionality of .skip-link is provided by .sr-only, which itself is partly overwritten by .skip-link.


pattonwebz:
This makes sense to me, should the `.skip-link` class also be removed inside the `header.php`?

IanDelMar:
@pattonwebz Thanks to your comment I had a look at this PR again and noticed the old story of removeing/re-structuring/ etc. may break child themes which do not have the `.sr-only` in their `header.php` yet.  So I elaborated on the PR.  I introduced a `deprecated.scss` and put the `.skip-link` there. I think it's now safe to remove `.skip-link` class from the header unless a child theme uses it for JS stuff. However, I'd suggest to have an `#skip-link` id for those who want to target this element by JS.

IanDelMar:
`.skip-link` is the default class used by `skip-link-focus.js` (see https://github.com/cedaro/skip-link-focus) - so closing this.

IanDelMar:
Was looking at the wrong repo. UnderStrap is using the undescores-for-npm packages for this. `skip-link-focus-fix.js` does not rely on any class. Reopen...

0dp:
When I was doing the works on the gulp4 or #883 I noticed that it's really not a lot of stuff from underscores that is being used. 
Skip-link and media seems to be the only two things being used from the npm package.
Correct me if I am wrong, but do we really need underscores as a dependency for the theme?

PR #898, opened 2019-01-23
Child Theme Support for `/css/theme.min.css`

craigiswayne:
When this theme is used as a parent theme, the link to `theme.min.css` references a uri in the child theme. 
Unless one is created, it will result it broken styles.

.... side note: maybe if you want the file to be overridden via the child theme we could use `get_theme_file_uri()` instead?:

```php
wp_enqueue_style( 'understrap-styles', get_theme_file_uri('/css/theme.min.css'), array(), $css_version );
```


craigiswayne:
Duplicate of #689

PR #967, opened 2019-05-01
Add support for wp_body_open

IanDelMar:
5.2 will introduce a new `wp_body_open()` function that is used to trigger a `wp_body_open` action. This action is intended to allow developers to inject code immediately following the opening `<body>` tag. Themes are encouraged to begin using this hook as soon as 5.2 is released.

See https://make.wordpress.org/core/2019/04/24/miscellaneous-developer-updates-in-5-2/

### DIFF
--- a/header.php
+++ b/header.php
@@ -23,7 +23,13 @@ $container = get_theme_mod( 'understrap_container_type' );
 </head>
 
 <body <?php body_class(); ?>>
-
+<?php 
+if ( function_exists( 'wp_body_open' ) ) {
+    wp_body_open();
+} else {
+    do_action( 'wp_body_open' );
+}
+?>
 <div class="site" id="page">
 
 	<!-- ******************* The Navbar Area ******************* -->


pattonwebz:
Just noting here that on my original post about this there is a comment to a ticket about handling of back-compat https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/#comment-43716. It was decided that it is enough for plugins to detect usage of the hook (with a `did_action()` call) after it should have fired and run their fallbacks if the hook has not yet fired.

Also for the sake of code simplicity here it is sufficient to call the `do_action()` directly instead of the wrapper as noted by Justin here: https://make.wordpress.org/themes/2019/03/29/addition-of-new-wp_body_open-hook/#comment-43714.

Also big +1 from me for adding this now :)

IanDelMar:
@pattonwebz Thank you! I removed the wrapper.

Issue #976, opened 2019-05-17
Woocommerce 3.6.3 Plugin could not be activated because it triggered a fatal error.


jpmorandesign:
- [X] a Bug Report
  - **WordPress Version:** 5.2
       WooCommerce 3.6.3 update

Upon updating to WoCommerce 3.6.3 I got the following error message and WooCommerce was deactivated

**Plugin could not be activated because it triggered a fatal error.**
Fatal error: Cannot redeclare wc_reviews_enabled() (previously declared in /nas/content/live/bc4less/wp-content/themes/understrap/inc/woocommerce.php:151) in /nas/content/live/bc4less/wp-content/plugins/woocommerce/includes/wc-conditional-functions.php on line 430
![Screen Shot 1](https://user-images.githubusercontent.com/17971314/57960874-844de300-78c8-11e9-8dfa-6b60746d8a6e.png)

IanDelMar:
#961

Thomas-A-Reinert:
Is this fixed? Realibly and confirmed? Would like to close the issue if it´s in the _official / released_ codebase.

jpmorandesign:
I have updated the file and WooCommerce wasn't automatically disabled, so I think it is resolved

Thomas-A-Reinert:
Thanks for your feedback, @jpmorandesign ! Appreciated!

PR #977, opened 2019-05-18
Fix for #976

IanDelMar:
Declare `wc_reviews_enabled()` and `wc_reviews_enabled()` on frontend only.

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -139,7 +139,7 @@ if ( ! function_exists( 'understrap_wc_form_field_args' ) ) {
 	}
 }
 
-if ( ! function_exists( 'wc_review_ratings_enabled' ) ) {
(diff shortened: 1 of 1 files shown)


IanDelMar:
@holger1411 I think this fix is important

Beaniie:
> Declare `wc_reviews_enabled()` and `wc_reviews_enabled()` on frontend only.

Could you elaborate? Where would these need to be declared?

Regards, 
-B.

IanDelMar:
In `inc/woocommerce.php` replace
```php
if ( ! function_exists( 'wc_review_ratings_enabled' ) ) {
```
by 
```php
if ( ! is_admin() && ! function_exists( 'wc_review_ratings_enabled' ) ) {
```

pattonwebz:
This looks good to me. Activated the theme (with this PR code change), activated Woocommerce. I get no fatal error :)

PR #990, opened 2019-06-20
Use shorter ABSPATH check

ylkyrg:


IanDelMar:
WordPress Coding Standards: "readability is more important than cleverness or brevity". See https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code.  That's why I used the longer version.

ylkyrg:
I think the comment covers it for readability. This is not something that anyone will be interacting with and it's not common to find it in themes so if it's going to be included at all I think making it as unobtrusive as possible is the way to go.

https://github.com/Automattic/_s/pull/1196#issuecomment-329699476

PR #1016, opened 2019-07-23
Wpcs i18n string literal

ylkyrg:
Translation functions require string literals.

The `content-none.php` template does not need to be nested in the `<ul>` tag.

### DIFF
--- a/author.php
+++ b/author.php
@@ -61,35 +61,35 @@ $container = get_theme_mod( 'understrap_container_type' );
 
 				</header><!-- .page-header -->
 
-				<ul>
+				<!-- The Loop -->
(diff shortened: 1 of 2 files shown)


Thomas-A-Reinert:
This seems odd on first sight. Can you explain why you removed the 'understrap' translation domain name?

pattonwebz:
Looks like it was removed because it's not yranslatable (translations are done with static analysis). 

Good catch :)

Thomas-A-Reinert:
Oh :man_facepalming:
Sure. Didn´t spot that on first sight. Looks totally okay now 👍

PR #1027, opened 2019-08-21
Custom Editor Styles don't get minified

sol-idsnake:
Added custom-editor-style to minify task in gulpfile.js


Thomas-A-Reinert:
Can´t confirm that this is working now, but CAN confirm that there never has been any task or even empty file as a placeholder that was meant to add custom-editor-styles.

IanDelMar:
Has been added in PR #1040

IanDelMar:
@UnderstrapFramework Fixed in #1040 -> close.

PR #1038, opened 2019-09-25
Update gulpconfig.json to make watch-bs task work

sol-idsnake:
Browser-sync expects an object as its init() config. browserSyncWatchFiles is currently an array, but has to be an object to include "files."

### DIFF
--- a/gulpconfig.json
+++ b/gulpconfig.json
@@ -3,7 +3,9 @@
 		"proxy": "localhost:8080/",
 		"notify": false
 	},
-	"browserSyncWatchFiles": ["./css/*.min.css", "./js/*.min.js", "./**/*.php"],
+	"browserSyncWatchFiles": {
+		"files": ["./css/*.min.css", "./js/*.min.js", "./**/*.php"]
+	},
 	"paths": {
 		"js": "./js",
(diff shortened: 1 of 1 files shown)


phoshenwen:
@alexbogias Some more info would be helpful.. What error message do you get?

Thomas-A-Reinert:
Needs testing. Can anyone confirm it didn´t work before and does now?
Worked for me since ages, so that´s why I´m asking.

alexbogias:
Looks ok for me now...

IanDelMar:
I haven't tested it, but the documentation supports this PR:
```js
// single file
browserSync({
    files: "app/css/style.css"
});

// multiple files
browserSync({
    files: ["app/css/style.css", "app/js/*.js"]
});
```
See https://www.browsersync.io/docs/options/#option-files

IanDelMar:
@sol-idsnake While the current watch-bs task works for me, your PR doesn't. I guess the reason is that `.init()` expects _one_ optional object and _one_ optional callback (https://www.browsersync.io/docs/api#api-init). Your PR provides _two_ objects as config for `.init()`.

PR #1063, opened 2019-11-12
Add aria-label to <nav>

IanDelMar:


Thomas-A-Reinert:
Hey @IanDelMar - your PR is well-intentioned, but _totally_ missing the point. In short: aria labels should be "meaningful". Try to get into the role of a blind user that gets told "MAIN". Would that be meaningful to you? How about "Main Navigation"? I assume you tried to stick with the strings that are already available so far, but that won´t work here. We need to upgrade that. Not only because of the a11y thing, but..

Thomas-A-Reinert:
In case you change that to a new "meaningful" string i´m going to merge that asap. In case I still got the privs. But that´s a thing you gotta ask @holger1411 or one of his bot minions.

IanDelMar:
If I am _totally_ missing the point, someone should tell Mozilla!

"A brief description of the purpose of the navigation, omitting the term "navigation", as the screen reader will read both the role and the contents of the label."
[Source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role)

Thomas-A-Reinert:
Sorry @IanDelMar : no offense intended, but that's what I have been taught. I'm off work this week due to being sick, but I will check my guru next week, okay?

Thomas-A-Reinert:
Sidenote: she plays a major role in Germany concerning accessibility and actually teaching me a good bit so guess I gonna believe her.

IanDelMar:
Do that, but I don't think she's gonna tell you anything else. All the sources I have read say that the landmark roles are read and give examples for aria-label without the word "navigation".

However, get well soon.

Thomas-A-Reinert:
Thanks. Just had an interesting discussion with her.
First: You did nothing wrong, it´s absolutely correct so far. On contrary: You really should NOT use the word navigation inside a navs aria label as you said!

Problem may occur using that solution if the term gets translated. In German "Main" would probably be translated into "Haupt" which could also be interpreted as "head in german language". Don´t know which other confusions may occur with other languages.

Anyway, there´s a margin for improvements.

We usually handle it differently by using semantic markup instead of aria labels wherever possible. As this is a sectioning element it should have a headline anyway and you can drop the aria label therefore.

The way we´d handle that would look something like that:
```
<nav class="navbar navbar-expand-md navbar-dark bg-primary">
  <h2><?php echo esc_attr_x( 'Main Navigation', 'navigation', 'understrap' ); ?></h2>
```

Or maybe even better: It could/should maybe take the name of the menu given by the admin. Problem there: S/he may "forget" about the "Navigation" term. So that´s a certain twist here.

IanDelMar:
> The way we´d handle that would look something like that:
> 
> ```
> <nav class="navbar navbar-expand-md navbar-dark bg-primary">
>   <h2><?php echo esc_attr_x( 'Main Navigation', 'navigation', 'understrap' ); ?></h2>
> ```
> 
> Or maybe even better: It could/should maybe take the name of the menu given by the admin. Problem there: S/he may "forget" about the "Navigation" term. So that´s a certain twist here.

I doubt that any German-speaking user would interpret "Haupt navigation" as "Kopf navigation" but, well, who knows... I definitely like the solution of giving the navigation a heading. I'd go for the fixed string version which may be changed by the developer or the WordPress familiar admin to avoid problems with certain naming habits.

So it'd become
```php
<nav class="navbar navbar-expand-md navbar-dark bg-primary">
  <h2 class="sr-only">
    <?php echo wp_strip_all_tags( __(  'Main Navigation', 'understrap'  ) ); ?>
  </h2>
```

What do you think?

Thomas-A-Reinert:
Well, combined with the term "navigation" - no way. Alone, well "maybe" yes. Anyway, this was just an example. Your example seems fine to me. I forgot about the sr-only ;)

Thomas-A-Reinert:
@IanDelMar ? I feel sorry, but my WCAG-Sensei doesn´t really reveal everything in the first run. She stacks up one on the other. Love it or hate it. Didactic I think it´s quite okay. But that means that I still got an improvement to suggest.. :/

```HTML
<nav class="navbar navbar-expand-md navbar-dark bg-primary" aria-labelledby="mainNavLabel">
  <h2 class="sr-only" id="mainNavLabel">
    <?php esc_html_e( 'Main Navigation', 'understrap' ); ?>
  </h2>
.. 
</nav>
```

That way you can totally distribute both over the whole document without any context and regardless "connect" them. This is comparable to the label tag and it´s "for"-attribute.

Do you want to add up to that as it´s not merged anyway or should I come up with a final (?) solution?

IanDelMar:
Ah, completely forgot about aria-labelledby.

PR #1082, opened 2019-12-14
Update header.php

Thomas-A-Reinert:
Using semantically correct header instead of div, correcting indentation, removing useless blank lines

### DIFF
--- a/header.php
+++ b/header.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * The header for our theme
+ * The header for Theme UnderStrap
  *
- * Displays all of the <head> section and everything up till <div id="content">
+ * Displays all of the <header> section and everything up till <div id="content">
  *
  * @package understrap
  */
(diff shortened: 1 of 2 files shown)


header.php
```diff
@@ -25,30 +25,27 @@
...
 		<nav id="main-nav" class="navbar navbar-expand-md navbar-dark bg-primary" aria-labelledby="main-nav-label">
-
 			<h2 id="main-nav-label" class="sr-only">
 				<?php esc_html_e( 'Main Navigation', 'understrap' ); ?>
 			</h2>
 
-		<?php if ( 'container' === $container ) : ?>
-			<div class="container">
-		<?php endif; ?>
+      <?php if ( 'container' === $container ) : ?>
```
IanDelMar: I think there should be 2 more tabs

Thomas-A-Reinert:
Yup. Did this without my own env on my ladies computer when I couldn´t sleep ;)
Bad preconditions for a perfect update.. Will do that again somewhen within the next week.  footer.php shouldn´t have been in there too, had to set up the whole process again. It just felt wrong not working pon my own dev-env :P

Or feel free to do it yourself. I´m not really envious about any merge/pull request that gets the inspiration from me. Only thing that counts is a – hopefully – perfect release ;)

PR #1102, opened 2019-12-20
Update dependencies (composer)

IanDelMar:


ylkyrg:
The lock file should also be updated with `composer update`.

Note that in this case specific versions do not need to be changed in `composer.json`. The caret symbol constrains a dependency to a particular major version. From the [docs](https://getcomposer.org/doc/articles/versions.md#caret-version-range-):  `^1.2.3` is equivalent to `>=1.2.3 <2.0.0`.

So with the version `^1.2.3`, running `composer update <package>` will update the [lock file](https://getcomposer.org/doc/01-basic-usage.md#installing-with-composer-lock) if there is a new version, provided it is not a major release.

IanDelMar:
See #1149

PR #1103, opened 2019-12-23
Remove task 'cssnano'

IanDelMar:
The task is a relic. `gulp-cssnano` has been removed in commit 1208774.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -87,28 +87,6 @@ gulp.task(
 	})
 );
 
-// Run:
-// gulp cssnano
-// Minifies CSS files
-gulp.task('cssnano', function () {
-	return gulp
-		.src(paths.css + '/theme.css')
-		.pipe(sourcemaps.init({ loadMaps: true }))
-		.pipe(
-			plumber({
-				errorHandler: function (err) {
-					console.log(err);
-					this.emit('end');
-				}
-			})
-		)
-		.pipe(rename({ suffix: '.min' }))
-		.pipe(cssnano({ discardComments: { removeAll: true } }))
-		.pipe(sourcemaps.write('./'))
-		.pipe(gulp.dest(paths.css))
-		.pipe(touch());
-});
-
 gulp.task('minifycss', function () {
 
 	return gulp


0dp:
Word!

I was thinking about going through the gulp setup this holiday, update dependencies and do a little tidy up.
Still think we can do a lot better in terms of features and performance.
I also think there is some merits in updating the file structure a bit similar to what I proposed last year #883 
Move all assets to src, drop copy-assets and just load node-modules directly, distinct directory for modules,. @UnderstrapFramework maybe it's something to consider and take a chat about again for a future roadmap?

UnderstrapFramework:
That would be great

IanDelMar:
@0dp I already started doing that (update dependencies, remove dependencies no longer in use, check whether we can do major version updates) on a local install.

0dp:
Feel free to have a look at my previous pr if there's anything useful 😉

IanDelMar:
I should have looked there first. :-D Why did this never make it into the master? While I cannot contribute in terms of features and performance, but I can test and check the dependencies.

0dp:
I have no idea?
It had all the new goodies at the time, and enabled us to have a small yet powerful build system, with lots of room for further optimization.
Yeah could be cool to see if it still holds up. At least there's a few things we can cherry pick

Thomas-A-Reinert:
It never got committed because holger blew it TWICE to merge it. And alongside the second PR I BEGGED him to get this thingie merged, because I wanted to get your effort being honored. To no avail. Sorry.

0dp:
If there is support for it, it shouldn't take too long to rework a new PR.

IanDelMar:
Started to build upon your PR here https://github.com/IanDelMar/understrap/tree/gulp-workflow

0dp:
> Started to build upon your PR here https://github.com/IanDelMar/understrap/tree/gulp-workflow

Sick dude!

Let me know if you need anything from me

0dp:
I wish I could comment in your repo.

There is some very noticeable moves we can make to streamline the build features.

For one,
We don't need to copy-assets task
We can load the the module files directly where needed. Which means we can just have to run the build task when updating modules. Instead of now where we need to run copy-assets and then build task to upgrade.

Secondly, I proposed that we copy the skip-link fix and media file from underscores and include them in the theme natively. They havnt been updated for ages in the original package. And since we only need those two files i think its overkill to include the whole underscores theme.
(Do we even need those files?)

IanDelMar:
Give me 1-3 days and I'll make a PR for discussion and further improvements. 
@ 2.: I already dropped the [underscores-for-npm dependency](https://github.com/understrap/understrap/commit/d73bfbcb162d3f263785a13839a974fe4f5ca975)

0dp:
Incremental builds

Bruners:
I have tested your gulp-workflow and it seems nice, but when i run gulp build i does not minify the js and css files. I also see some webpack stuff added in the scripts, is that intended?

https://gist.github.com/Bruners/86feb06183a3a349b1665af4a928e13c

IanDelMar:
Hey! Now the 1-3 days are turning to almost 2 months. I'm sorry for the delay but I had and have a lot of work in my real job. I think I won't be able to finish before March.

@Bruners I haven't pushed all commits yet. I switched to use rollup instead of webpack and started to break down all tasks into smaller tasks (e.g. a compile SCSS to CSS task for the theme.css and the custom-admin-styles.css) to utilize parallelization. I'll try to do a little cleanup and push those commits within 3-4 weeks. It seems that the theme got stuck again, so there's no hurry anyways.

Thomas-A-Reinert:
Hey @IanDelMar - personally i´m totally fine with rollup.js. But don´t you think people will maybe feel a bit uncomfy with it because it´s not as "established" as webpack, browserify or parcel? Just saying..

Is there a reason for the switch? Would you be so kind to share some insights about the reasons?

IanDelMar:
@Thomas-A-Reinert I found it easier to set up and it performed better in terms of build times.

PR #1106, opened 2019-12-27
Update understrap_post_nav()

IanDelMar:
* Moves the template tag from `inc/extras.php` to `inc/template-tags.php`.
* Removes the translations which are probably there to change the Font Awesome stuff which should not be up to the translator.
* Improves accessibility by adding the attributes `aria-labelledby` and `aria-hidden`.
* Styles the post navigation with Bootstrap's `.nav` and `.nav-link`.

@ Font Awesome: AFAIK this is the only place where the theme uses FA. [theme.scss](https://github.com/understrap/understrap/blob/master/sass/theme.scss) states that the inclusion of FA is optional which it is not unless one enqueues FA in any other way. 
**Needs discussion**: do we really need to use FA there? Wouldn't it be better to provide something like
```scss
.post-navigation .nav-previous { @include caret(left); }
.post-navigation .nav-next { @include caret(right); }
````
to make the exclusion of FA safe by default?

@Thomas-A-Reinert could you please check the accessibility part? Thank you!

_NOTE_: The re-styling part would induce a major version increment. It could break a child theme's styling.

### DIFF
--- a/inc/extras.php
+++ b/inc/extras.php
@@ -222,3 +222,25 @@ if ( ! function_exists( 'understrap_kses_title' ) ) {
 		return wp_kses( $data, $allowed_tags );
 	}
 } // End of if function_exists( 'understrap_kses_title' ).
+if ( ! function_exists( 'understrap_post_nav_link_class' ) ) {
+	/**
+	 * Adds CSS class to a post navigation link.
+	 *
+	 * @param string  $output   HTML markup for the adjacent post link..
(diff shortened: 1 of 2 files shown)


Thomas-A-Reinert:
@IanDelMar : Get rid of FA. It adds bloat and really just gets used where you mentioned it. Icons can easily be added via UTF-8-Codes (not accessible) and or inline-svg (WCAG compliant if done right) if needed.

PR #1124, opened 2020-02-27
Jetpack

gillespieza:
Added the id of the sidebar "footerfull" to the infinite-scroll argument list, so that if there are any widgets in the footer, the default Infinite Scroll behaviour would change to on-click instead.

Added a scss file to make the default layout of the Infinite Scroll button and footer match Bootstrap 4.

### DIFF
--- a/sass/theme.scss
+++ b/sass/theme.scss
@@ -1,12 +1,13 @@
-@import "theme/theme_variables";  // <-------- Add your variables into this file. Also add variables to overwrite Bootstrap or UnderStrap variables here
-@import "assets/bootstrap4";      // <-------- Loads Bootstrap4
-@import "understrap/understrap";  // <-------- Loads the UnderStrap defaults. Just a few classes to incorporate BS in WP
+@import "theme/theme_variables"; // <-------- Add your variables into this file. Also add variables to overwrite Bootstrap or UnderStrap variables here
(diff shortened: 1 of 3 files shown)


sass/theme.scss
```diff
@@ -1,12 +1,13 @@
...
 
 //Optional files - If you dont use the corresponding scripts/fonts comment em out
 @import "assets/font-awesome"; // <------- Font Awesome Icon font
 @import "assets/underscores"; // <------- Underscores media styles
 
 // Any additional imported files //
-@import "theme/theme";  // <--------- That's where you can add your own design. Thats your part!
+@import "theme/theme"; // <--------- That's where you can add your own design. Thats your part!
 @import "theme/contact-form7"; // Contact Form 7 - Bootstrap 4 support
+@import "theme/infinite-scroll"; // Infinite Scroll from JetPack.
```
IanDelMar: File `sass/theme/_infinite-scroll.scss` is missing. The import of non-existing files will throw an error on compilation.

gillespieza: Ooops! I've added the missing file.

danemorgan:
As it's been more than a year and a half, and we have conflicts here, I wonder if closing this pull request and making a fresh go of it starting from the current code would be a better approach?

danemorgan:
Closing as stale.

PR #1126, opened 2020-02-28
Sticky posts should be distinctly recognisable

gillespieza:
I've added some SCSS to the bottom of understrap.scss to highlight sticky posts. They should be distinct from regular posts in a list of posts. This adds bg-light, a left border and a bit of padding to posts with the class .sticky.

### DIFF
--- a/sass/understrap/understrap.scss
+++ b/sass/understrap/understrap.scss
@@ -1,54 +1,71 @@
 // Some basic padding for all wrappers
 .wrapper {
-  padding: $grid-gutter-width 0; }
+	padding: $grid-gutter-width 0;
+}
 
 // Reset hero wrapper padding to 0
-#wrapper-hero { padding: 0px !important; }
+#wrapper-hero {
+	padding: 0px !important;
+}
 
 // Adding basic WordPress classes to pass the WordPress.org tests
 .sticky,
 .gallery-caption,
 .bypostauthor {
-  font-size: inherit;
+	font-size: inherit;
 }
 
 // Separate sticky wrapper from main content
-.wrapper#wrapper-sticky { border-bottom: 1px solid $gray-300; }
+.wrapper#wrapper-sticky {
+	border-bottom: 1px solid $gray-300;
+}
 
 // Adding some contrast background color to footer full widget
 #wrapper-footer-full,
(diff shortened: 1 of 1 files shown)


IanDelMar:
> They should be distinct from regular posts in a list of posts.

They "can be" but I do not see a they "should be".

gillespieza:
In the `themeunittestdata.wordpress.xml` file from WordPress.com: (https://wpthemetestdata.wordpress.com/2012/01/07/template-sticky/)

> The sticky post should be distinctly recognizable in some way in comparison to normal posts. You can style the .sticky class if you are using the post_class() function to generate your post classes, which is a best practice.

Thomas-A-Reinert:
I totally agree. Anyway, as this is a starter theme I´d suggest leaving the styling to the dev using it, or not?

gillespieza:
If someone wants to jazz it up, great, they could absolutely do that in their child theme or in theme.scss. All these changes do is allow a person to use Understrap straight out of the box, without having to muck around with adding additional CSS in the customiser, etc. It is very minimal styling - no more intrusive than the basic styling of the footer, or the navbars, or the Hero widgets.

But maybe I'm missing the point? Is the goal of Understrap, as a starter theme, to be a complete (if minimalist) theme that someone could use out-the-box, or should users _always_ be creating child themes and continuing from there? If the latter, then sure, leave the styling to them. Are there guidelines on what styling belongs to the user/child theme and what should be included in the parent theme? I couldn't find any milestones or roadmaps or current contribution guidelines...

0dp:
The latter would be the case for this theme.

Regarding the guidelines and roadmaps. This is something we are all missing. We've been requesting this ever since the repo owner stepped down and put a new one in charge

IanDelMar:
@gillespieza Now I do see the "should" :-) I checked the [theme handbook](https://developer.wordpress.org/themes/functionality/sticky-posts/) only and there it reads more like a "can". Thank you for the reminder to also check the theme unit test!

Thomas-A-Reinert:
There is some kind of validator included in [Mythic](https://github.com/justintadlock/mythic) for example that can also run tests versus the repo guidelines.

PR #1139, opened 2020-03-22
update quantity-input.php template

tvanro:
placeholder support has been added in the latest official 4.0.0 version https://github.com/woocommerce/woocommerce/blob/master/templates/global/quantity-input.php

### DIFF
--- a/woocommerce/global/quantity-input.php
+++ b/woocommerce/global/quantity-input.php
@@ -12,7 +12,7 @@
  *
  * @see     https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce/Templates
- * @version 3.6.1
+ * @version 4.0.0
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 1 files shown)


noelspringer:
@tvanro as mentioned here https://github.com/understrap/understrap/issues/1138#issuecomment-602971373 this file doesn't provide any Bootstrap specific visual display. The Bootstrap class `sr-only` replaces the default Woocoomerce class `screen-reader-text`. The default Woocommerce class is perfectly adequate and the simple solution is to remove the file `quantity-input.php` altogether from UnderStrap.

Cheers

tvanro:
Ok, I see. Then maybe it could be more useful to just delete `quantity-input.php` from this repository instead of updating it... 

Thanks for the insights @noelspringer !

lukaspetr:
Hi @tvanro , the wasn't deleted as far as I see it. Is that a final decision?

PR #1140, opened 2020-03-30
Automate converting styles to RTL

kfathy194:
**Automate converting styles into RTL by adding the following:**

- Added gulp-rtlcss to package.json dependencies.
- Added require gulp-rtlcss in gulpfile.js
- Added pipe for rtlCSS() in the sass task
- Added regex to look for theme.css or theme-rtl.css in the cssnano and minifycss tasks.
- Added if/else for enqueue styles in the enqueue.php.


danemorgan:
I think I prefer the approach in [#1250](https://github.com/understrap/understrap/pull/1250). Recommend closing this pull request.

PR #1155, opened 2020-04-18
Update comments

IanDelMar:
* The HTML for the comment form fields is outdated. Using the proposed str_replace updates the HTML to WP's default HTML and makes it easier to maintain in the future.
* Use the comment_form_comments_closed to display the notice that comments are closed.
* Minor WPCS fixes.

### DIFF
--- a/comments.php
+++ b/comments.php
@@ -49,7 +49,7 @@ if ( post_password_required() ) {
 							'understrap'
 						)
 					),
-					number_format_i18n( '$comments_number' ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					number_format_i18n( $comments_number ), // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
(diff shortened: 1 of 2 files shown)


IanDelMar:
@UnderstrapFramework should be safe to merge. However, maybe I'm just not aware of a case breaking child themes.

UnderstrapFramework:
Maybe i'm missing something but the change in this:

number_format_i18n( $comments_number )

to this:

number_format_i18n( '$comments_number' )

wont work? I'm aware that double quotes will process the variable but not single? @IanDelMar

IanDelMar:
Gosh, @UnderstrapFramework you're absolutely right! I fixed it in 
092e4cb.

UnderstrapFramework:
I'm the human PHP debugger! Thanks for sorting it @IanDelMar

Thomas-A-Reinert:
This is ridiculous.

PR #1165, opened 2020-05-03
Update Travis

IanDelMar:
Builds were failing due to outdated travis.myl

* Adds php-parallel-lint and check for PHP parse errors.
* Adds WPThemeReview rule ~~, script and check~~.
* ~~Adds PHPcompatibilityWP script and check.~~
* Removes install of WordPress.


.travis.yml
```diff
@@ -4,94 +4,62 @@
...
-    # -p flag: Show progress of the run.
-    # -s flag: Show sniff codes in all reports.
-    # -v flag: Print verbose output.
-    # -n flag: Do not print warnings (shortcut for --warning-severity=0)
-    # --standard: Use WordPress as the standard.
-    # --extensions: Only sniff PHP files.
-    - $PHPCS_DIR/bin/phpcs -p -s -v -n . --extensions=php --ignore=*/woocommerce/*,*/src/*
+  # Validate the composer.json file.
+  # @link https://getcomposer.org/doc/03-cli.md#validate
+  - composer validate --no-check-all --strict
```
jrfnl, line 55: `composer install` already does a basic validation, so you don't really need to run this command on all PHP versions. Running it on one should be enough.

composer.json
```diff
@@ -15,13 +15,18 @@
 		"php": ">=5.6"
 	},
 	"require-dev": {
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.6",
 		"roave/security-advisories": "dev-master",
-		"wptrt/wpthemereview": "^0.2"
+		"wptrt/wpthemereview": "*",
+		"dealerdirect/phpcodesniffer-composer-installer": "0.6",
```
jrfnl, line 20: Please use the caret to get the latest patch version of the tool. Versions 0.6.0 and 0.6.1 have known bugs. 0.6.2 is the latest and will be used automatically when using the caret for the versioning.

```suggestion
		"dealerdirect/phpcodesniffer-composer-installer": "^0.6",
```

composer.json
```diff
@@ -15,13 +15,18 @@
...
+		"php-parallel-lint/php-parallel-lint": "^1.2"
 	},
 	"scripts": {
-		"check-cs": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -w"],
-		"fix-cs": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"]
+		"check:cs": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -p"],
+		"check:compatibility": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -p --standard=phpcs.xml,PHPCompatibilityWP"],
+		"check:theme-review": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcs -p --standard=phpcs.xml,WPThemeReview"],
```
jrfnl, line 26: These don't make any sense as they will just result in the same as `check:cs` as both rulesets are already included in the `phpcs.xml` file.

composer.json
```diff
@@ -15,13 +15,18 @@
...
+		"lint:php": ["@php ./vendor/bin/parallel-lint --exclude .git --exclude vendor ."],
+		"checks": [ "@check:cs", "@check:compatibility", "@check:theme-review"],
```
jrfnl, line 28: Aside from my above remark, I have bad experienced with combining checks like this with automated build checking as the error codes are not picked up correctly, so the build will always pass, even when checks are failing.

composer.json
```diff
@@ -15,13 +15,18 @@
...
+		"fix:cs": ["@php ./vendor/squizlabs/php_codesniffer/bin/phpcbf"]
```
jrfnl, line 29: Just from a sanity perspective, you may want to keep the `fix:cs` and `check:cs` together (without the unrelated `lint:php` in between them).

phpcs.xml
```diff
@@ -86,7 +86,8 @@
 
 	<!-- Use the PHPCompatibility Ruleset -->
 	<config name="testVersion" value="5.6-99.0"/>
-	<rule ref="PHPCompatibilityWP">
-		<include-pattern>*\.php$</include-pattern>
-	</rule>
-</ruleset>
+	<rule ref="PHPCompatibilityWP" />
```
jrfnl, line 89: Just wondering why the `<include-pattern>` is being removed.

I did see that the extensions are limited to `php`, but considering WPThemeReview also contains some non-php checks, removing that and leaving this would make more sense.

.travis.yml
```diff
@@ -4,94 +4,62 @@
...
+# Declare project language and PHP versions to test against.
+# @link https://docs.travis-ci.com/user/languages/php/
+language: php
+php:
+  - '5.6'
+  - '7.0'
+  - '7.1'
+  - '7.2'
+  - '7.3'
+  - '7.4'
```
jrfnl, line 27: You may want to add a build against PHP `nightly` (= PHP 8) which is allowed to fail to use as an early warning system for the upcoming changes.

This may need some tweaking as I don't think the PHPCS dependencies will install on PHP 8 yet, though Parallel-Lint should be ok.

composer.json
```diff
@@ -15,13 +15,18 @@
 		"php": ">=5.6"
 	},
 	"require-dev": {
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.6",
```
jrfnl, line 18: You seem to be missing `wp-coding-standards/wpcs`.

I realize it's pulled in via `WPThemeReview` too, but it is also an explicit dependency of your toolset as you include the `WordPress` ruleset in your `phpcs.xml` file.

composer.json
IanDelMar: Thank you! The caret got lost due to copy and paste.

.travis.yml
IanDelMar: Thank you!

.travis.yml
IanDelMar: You're right! I first included `nightly` but removed it as the `composer install` was throwing an error. I'll think about a tweak.

composer.json
IanDelMar: Thanks for the hint!

composer.json
IanDelMar: You're right. Thank you!

composer.json
IanDelMar: Thank you!

phpcs.xml.dist
```diff
@@ -13,7 +13,7 @@
 	<arg name="basepath" value="./"/>
 
 	<!-- Show colors in console -->
-	<arg value="-colors"/>
+	<arg value="colors"/>
```
jrfnl, line 16:
```suggestion
	<arg name="colors"/>
```

phpcs.xml.dist
```diff
@@ -13,7 +13,7 @@
 	<arg name="basepath" value="./"/>
```
jrfnl, line 13: As the file has now been renamed, you'll probably want to add `phpcs.xml` and `.phpcs.xml` (the names of the stnadard supported overload files) to the `.gitignore` file.

IanDelMar:
@jrfnl Thank you very much for taking the time to go through the PR and makeing all the helpful comments and suggestions!

> * Rename the `phpcs.xml` file to `phpcs.xml.dist` to allow overloading of the file by devs for their local environment.

Good idea! Thank you!
> * Set an explicit file to use for the PHPCS cache.

What's the advantage of this?
> * To use colors in the PHPCS output, just use `<arg name="colors"/>`

Once I do that it results in the following error
image

> * Why include `Zend.Files.ClosingTag` ? WPCS already includes `PSR2.Files.ClosingTag` for the same.

Thank you!
> * `<config name="testVersion" value="5.6-99.0"/>` -> you can make this open ended by setting it like so: `<config name="testVersion" value="5.6-"/>`

Thank you!

phpcs.xml.dist
IanDelMar: Thank you!

jrfnl:
> > Set an explicit file to use for the PHPCS cache.
>
> What's the advantage of this?

See  https://github.com/squizlabs/PHP_CodeSniffer/issues/530. `--cache` and `--cache=<file>` behave differently. The former means the cache will not be invalidated if the standard changes, but the latter means it will. 

> > To use colors in the PHPCS output, just use <arg name="colors"/>
>
> Once I do that it results in the following error

That error is unrelated to this change. What else did you change ? Possibly something in the composer scripts ?

IanDelMar:
> That error is unrelated to this change. What else did you change ? Possibly something in the composer scripts ?

No, it is not. I had to change `<arg value="colors"/>` to `<arg name="colors"/>`. `<arg value="-colors"/>` does not throw an error. Neither does the correct version  `<arg name="colors"/>`. But  `<arg value="colors"/>` throws the error I mentioned above. Changing `<arg value="colors"/>` to `<arg name="colors"/>` fixed that.

jrfnl:
> No, it is not. I had to change <arg value="colors"/> to <arg name="colors"/>. <arg value="-colors"/> does not throw an error. Neither does the correct version <arg name="colors"/>. But <arg value="colors"/> throws the error I mentioned above. Changing <arg value="colors"/> to <arg name="colors"/> fixed that.

I seem to be missing something, but your "solution" is **exactly** the change _I suggested_.

IanDelMar:
> I seem to be missing something, but your "solution" is **exactly** the change _I suggested_.

Yep, you suggested this 1h ago. Before that it was just to remove the dash. I haven't seen your updated suggestion when removing the dash and writing my comment. Sorry that I was not precise enough in 
> No, it is not. I had to change <arg value="colors"/> to <arg name="colors"/>. <arg value="-colors"/> does not throw an error. Neither does the correct version <arg name="colors"/>. But <arg value="colors"/> throws the error I mentioned above. Changing <arg value="colors"/> to <arg name="colors"/> fixed that.

It was your updated suggestion that made me see what has to be changed and in the meantime I commited your suggestions.

IanDelMar:
> See [squizlabs/PHP_CodeSniffer#530](https://github.com/squizlabs/PHP_CodeSniffer/issues/530). `--cache` and `--cache=<file>` behave differently. The former means the cache will not be invalidated if the standard changes, but the latter means it will.

Ah, I see. Thank you for patiently explaining things to me!

composer.json
IanDelMar: Thank you!

jrfnl:
> Yep, you suggested this 1h ago. Before that it was just to remove the dash. 

I suggested the same both times :confused:

IanDelMar:
> > Yep, you suggested this 1h ago. Before that it was just to remove the dash.
> 
> I suggested the same both times 😕

Oh, so sorry! Again, you're right! I should have waited to really get awake before starting to work on this. I was not processing that your suggestion was not limited to remove the dash but also to change `value` to `name`.

.travis.yml
IanDelMar: Might this be a solution?
I added `-ignore-platform-reqs` for build against PHP `nightly`, so `composer install` does no longer throw an error because of PHPCS. This allows Parallel-Lint to run.
```
php:
  - '5.6'
  - '7.0'
  - '7.1'
  - '7.2'
  - '7.3'
  - '7.4'
  - nightly

matrix:
  fast_finish: true
  include:
    - php: 7.4
      env: PHPCS=1
  allow_failures:
    - php: nightly
  
# Use this to prepare the build for testing.
before_script:
    # Speed up build time by disabling Xdebug.
    - phpenv config-rm xdebug.ini || echo 'No xdebug config.'
    - if [ "$TRAVIS_PHP_VERSION" == "nightly" ]; then composer install --ignore-platform-reqs; fi
    - if [ "$TRAVIS_PHP_VERSION" != "nightly" ]; then composer install; fi
    # Install JSCS: JavaScript Code Style checker
    # @link http://jscs.info/
    - if [[ "$PHPCS" == 1 ]]; then npm install -g jscs; fi
    # Install JSHint, a JavaScript Code Quality Tool
    # @link https://jshint.com/docs/
    - if [[ "$PHPCS" == 1 ]]; then npm install -g jshint; fi
    - if [[ "$PHPCS" == 1 ]]; then wget https://develop.svn.wordpress.org/trunk/.jshintrc; fi
```

IanDelMar:
@UnderstrapFramework thanks to @jrfnl this PR is now ready to be merged. @jrfnl Thank you for the great support!

UnderstrapFramework:
Wow great work @jrfnl, thanks

PR #1186, opened 2020-06-01
Re-add PHPCS config file to ignored files in dist task

IanDelMar:
The PHPCS config file has been renamed from phpcs.xml to phpcs.xml.dist in #1165 without adjusting the list of file extensions which are ignored in the dist task. This PR adds .dist to that list of file extensions.

@UnderstrapFramework please merge


UnderstrapFramework:
Thanks @IanDelMar

PR #1187, opened 2020-06-01
Remove vendor from paths

IanDelMar:
Copying `skip-link-focus-fix.js` to `paths.js + paths.vendor` in the copy-assets task has been removed in https://github.com/understrap/understrap/commit/882cd196019268ac6635ec7f051c5344fc85a3ca (7 Apr 2018). No other file is copied to `paths.js + paths.vendor` either. Also, the string `'/js/skip-link-focus-fix.js'` does not occur in any file making the replacement obsolete.

This PR
- removes the replacement task
- removes paths.vendor from gulpconfig
- removes the dependency gulp-replace

@UnderstrapFramework please merge

### DIFF
--- a/gulpconfig.json
+++ b/gulpconfig.json
@@ -15,7 +15,6 @@
 		"composer": "./vendor",
 		"dev": "./src",
 		"dist": "./dist",
-		"distprod": "./dist-product",
-		"vendor": ""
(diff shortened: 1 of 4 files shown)


UnderstrapFramework:
Thanks @IanDelMar

PR #1188, opened 2020-06-02
Add rtl support

farzadfaal:
Adding a new file called rtl.css which is mentioned [here](https://github.com/understrap/understrap/#rtl-styles) will not work as it's supposed to. Because bootstrap (up to the current version) doesn't have native rtl support. So I implemented **gulp-rtlcss** in a task called **rtl** for mirroring every directional css property. I've tested the outcome which seems to [work fine](https://pasteboard.co/JbezMER.png).


IanDelMar:
Related to #1140 which incorporates the rtl conversion into the existing sass and minify tasks and also properly enqueues the rtl CSS file. @UnderstrapFramework I'd suggest to close this PR.

IanDelMar:
@UnderstrapFramework please close

PR #1189, opened 2020-06-04
Switch to case-sensitiv @package tag

IanDelMar:
@UnderstrapFramework safe to merge

### DIFF
--- a/404.php
+++ b/404.php
@@ -2,7 +2,7 @@
 /**
  * The template for displaying 404 pages (not found)
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
--- a/archive.php
+++ b/archive.php
@@ -4,7 +4,7 @@
  *
  * Learn more: http://codex.wordpress.org/Template_Hierarchy
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
--- a/author.php
+++ b/author.php
@@ -4,7 +4,7 @@
  *
  * Learn more: https://codex.wordpress.org/Author_Templates
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
--- a/comments.php
+++ b/comments.php
@@ -5,7 +5,7 @@
  * The area of the page that contains both current comments
  * and the comment form.
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
--- a/footer.php
+++ b/footer.php
@@ -4,7 +4,7 @@
  *
  * Contains the closing of the #content div and all content after
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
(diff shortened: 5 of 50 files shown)


UnderstrapFramework:
Thanks @IanDelMar

0dp:
But the package is called understrap all lowercase

IanDelMar:
@0dp which package are you referring to? It's quite common to have it not all lowercase for themes as well as for plugins. See e.g. the `@subpackage` tag of the Twenty Something themes or the `@package` tag for WooCommerce or Yoast SEO etc.. The `@package` tag ist for logical grouping and not for representing whatever package name you have in mind. The theme is called UnderStrap, so it makes perfectly sense to have `@package UnderStrap` which might then become e.g. `@package UnderStrap/Templates`, `@package UnderStrap/Includes` or whatever.

0dp:
Hey @IanDelMar I am referring to understrap/understrap in the repo name.
I would under normal circumstances consider the repo name and package name to match

IanDelMar:
@0dp, still the `@package` tag and the repo name are not related to each another in terms of functionality. The package the `@package` tag refers to the theme not to the git repo. The theme's name is UnderStrap not understrap. In fact, for installs via the WordPress theme repo the git repo and it's name have no relevance at all.

0dp:
I know they are not related. But keep it consistent that's all

IanDelMar:
Sorry, I still lack understanding for why it's (more) consistent to refer to the git repo rather than to the theme name. It was the most recent merged commit. So it's easy to just revert it in case the github repo name 'understrap' is more consistent than the actual theme name 'UnderStrap'.

0dp:
I'm not saying persay it is more consistent to refer to the GitHub repo
name for the package name et al.
Just that it's nice to keep the naming consistent. As to why you made this
pr, I presume.

Then maybe get the repo name updated so it matches the theme name

0dp:
Am I completely wrong on this one @IanDelMar , just say

IanDelMar:
@0dp that's excactly the question I'm asking myself :smiley:

For me it's consistent to let non development files refer to 'UnderStrap' as 
>In fact, for installs via the WordPress theme repo the git repo and it's name have no relevance at all.

We should't change the repo's name. This would be problematic on case-sensitive OS because `.../wp-content/theme/understrap` is not the same as  `.../wp-content/theme/UnderStrap`. The latter is what you get when clone https://github.com/understrap/understrap.git into `.../wp-content/theme`. When you download the zip file you geht `UnderStrap.zip` containing the folder `UnderStrap`. Most repositories use a lowercase name, thus most likely people are expecting a lowercase repo name. Also, ~100% of the people already using UnderStrap are expecting a lowercase repo name.

While changing the repos name affects functionality, changing the `@package` tag does not. It just increases consistency within the theme files and according to you introduces inconsistency regarding the repo name.

0dp:
Maybe I'm just too used to see the name in lowercase everywhere I look, that it seems out of place seeing it capitalized

IanDelMar:
@pattonwebz what do you think? I guess as part of the WP Themes team you have seen _a lot_ of  themes.

PR #1190, opened 2020-06-04
Remove Bootstrap 3 classes from form fields

IanDelMar:
- Removes Bootstrap 3 classes from form fields
- Switches to default from control size by not replacing .input-lg with .form-control-lg
- Fixes single line and multi-line comments according to WPCS (https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-inline-comments)

See https://getbootstrap.com/docs/4.5/migration/ for the most notable changes from v3.x to v4.x

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -2,7 +2,7 @@
 /**
  * Add WooCommerce support
  *
- * @package understrap
+ * @package UnderStrap
  */
 
 // Exit if accessed directly.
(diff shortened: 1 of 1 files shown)


IanDelMar:
@UnderstrapFramework No feedback so far and I think this is safe to merge.

pattonwebz:
No effect (except cleaner markup :smile_cat: ) from removing the old classnames. Slight design change by allowing default control size instead but personally I prefer that.

IanDelMar:
> No effect (except cleaner markup 😸 ) from removing the old classnames. Slight design change by allowing default control size instead but personally I prefer that.

The class `.input-lg` does not exist in Bootstrap 4 and thus adds no styles at all. Then it's no design change except someone used the Bootstrap 3 class to add styles, right?

pattonwebz:
I guess one of the plugins on my test setup has bootstrap 3 form classes that it outputs.

Tested again with a fresh install there are zero differences been original and this PRs changes :)

UnderstrapFramework:
Thanks everyone @pattonwebz @IanDelMar

Issue #586, opened 2018-03-04
Integrate Font Awesome 5 (with 4.7 option)


0dp:
> This post is subject to updates

Integrating Font Awesome into UnderStrap can be done in the following ways: 

## Manually, documented in the docs
Method of manual integration has been discussed and proved here #547 
This method should be documented regardless, in case a theme developer wants to change from FA4.7 to FA5 free or FA5 pro, or a developer wants to change from FA5 free to FA5 pro.

## Automated, via npm install script
NPM packages (package.json) supports `preinstall` scripts which will allow us to modify the package.json file before it is processed. ie. changing dependencies. Theoretically we should be able to setup the entire project according to #547 
Eventually, it could possibly also allow developers installing 3rd party components per project basis.
My only concern about this approach is that it does not seem to be a common way of configuring npm packages. available examples?

## Automated, via scaffolding systems like Yeoman (others?)

Yeoman would add an extra layer to the development process, but unlike npm preinstall scripts, it's made for scaffolding. 

```
What's Yeoman?

Yeoman helps you to kickstart new projects, prescribing best practices and tools to help you stay productive.

To do so, we provide a generator ecosystem. A generator is basically a plugin that can be run with the `yo` command to scaffold complete projects or useful parts.

Through our official Generators, we promote the "Yeoman workflow". This workflow is a robust and opinionated client-side stack, comprising tools and frameworks that can help developers quickly build beautiful web applications. We take care of providing everything needed to get started without any of the normal headaches associated with a manual setup.

With a modular architecture that can scale out of the box, we leverage the success and lessons learned from several open-source communities to ensure the stack developers use is as intelligent as possible.

As firm believers in good documentation and well thought out build processes, Yeoman includes support for linting, testing, minification and much more, so developers can focus on solutions rather than worrying about the little things.
```

There is already a generator available for underscores found [here](https://github.com/mnyorba/generator-npm-wp-s-theme) which could be used as template.

holger1411:
Puh, lot of opinions and ideas here: https://github.com/understrap/understrap/issues/547
I created some basic rules for UnderStrap here:
https://github.com/orgs/understrap/teams/understrap-playbook/discussions/1

That just a few things which are important to me. Everyone is free to comment this and I will add more rules or change/shape the current four (as long as we can define a small set of up to 7 main rules).

If I check these (current rules) we should:
- Remove any FA integration from .php files. Bootstrap uses another way to render small icons (like in the dropdown component or pagination):
[...]

PR #1199, opened 2020-06-17
Removed FontAwesome4 

JamesVanWaza:
Replaced it with FontAwesome5

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "understrap",
-  "version": "0.9.4",
+  "version": "0.9.5",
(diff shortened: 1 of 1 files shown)


IanDelMar:
Please see #547 and #586. Just changing the dependency won't work.

IanDelMar:
As simply changing the dependency without adjusting the gulp scripts and the corresponding classes in the php files and so on won't work and given the discussions mentioned above, I suggest to close this pull request. @JamesVanWaza feel free to come up with a more complete PR. @UnderstrapFramework please close.

PR #1200, opened 2020-06-17
Fix WooCommerce checkbox input

### LINKED ISSUE #964: Calling woocommerce_form_field for checkbox results in a hidden checkbox
## Issue is about..
  - [x] a Bug Report

  - **WordPress Version:** 5.1.x
  - **Browser:** all

---

## Steps to reproduce?
Add this code to any template:
```php
<?php woocommerce_form_field( 'test_checkbox',
    array(
        'type' => 'checkbox',
        'label' => 'Test',
    )
) ?>
```
Open a page rendered by the template.

End user

Custom fields added to checkout, etc

## *When* does this occur?
When a checkbox field is added to any WooCommerce field. We noticed the issue when enabled WooCommerce compatibility in GDPR Framework plugin, which added a consent checkbox to checkout form.

## *Where* on the platform does it happen?
understrap_wc_form_field_args function at inc/woocommerce.php:125 add custom-* Bootstrap classes that hide regular checkbox but do not show custom one

Call woocommerce_form_field function for a checkbox from any template or add a checkbox field to a WooCommerce form.

## Are there any related Issues and/or PRs?
Looks like radio buttons could be affected as well.

## Screenshots / Video
This is how checkbox fields look like:
image
image

## Expected behavior
Correctly shown checkbox either using regular or custom styles:
image

### Other Comments
Looks like understrap_wc_form_field_args function is intended to apply "custom-*" Bootstrap classes to WooCommerce fields but the resulting HTML markup for checkout is not the same as required by Bootstrap.

Markup generated:
```html
<p class="form-row " id="test_checkbox_field" data-priority="">
    <span class="woocommerce-input-wrapper">
        <label class="checkbox custom-control custom-checkbox">
            <input type="checkbox" class="input-checkbox custom-control-input input-lg" name="test_checkbox" id="test_checkbox" value="1"> Test&nbsp;<span class="optional">(optional)</span>
      </label>
    </span>
</p>
``` 

Markup suggested by Bootstrap 4 docs (https://getbootstrap.com/docs/4.1/components/forms/#custom-forms):
```html
<div class="custom-control custom-checkbox">
    <input type="checkbox" class="custom-control-input" id="customCheck1">
    <label class="custom-control-label" for="customCheck1">Check this custom checkbox</label>
</div>
```

However, for Bootstrap the order and nesting of tag is important. For example, this rule at _custom-forms.scss:22 assumes that label tag goes after the input and wouldn't work for input inside label like WooCommerce generates it:
```css
.custom-control-input {
  position: absolute;
  z-index: -1; // Put the input behind the label so it doesn't overlay text
  opacity: 0;

  &:checked ~ .custom-control-label::before {
    color: $custom-control-indicator-checked-color;
    @include gradient-bg($custom-control-indicator-checked-bg);
    @include box-shadow($custom-control-indicator-checked-box-shadow);
  }
...
```

As a temporary solution I overrode understrap_wc_form_field_args function in my child theme and commented out custom classes for checkbox:
```php
        case 'checkbox':
            //$args['label_class'] = array( 'custom-control custom-checkbox' );
            //$args['input_class'] = array( 'custom-control-input', 'input-lg' );
            break;
```

Either the theme could do the same to leave checkboxes (and possibly radio buttons ) with regular styles or implement "woocommerce_form_field_checkbox" filter to completely rewrite checkbox and radio field HTML up to Bootstrap expectations.

IanDelMar:
Problem: Calling `woocommerce_form_field()` for input type `checkbox` results in a hidden checkbox.

Fix: Adjust the `understrap_wc_form_field_args` filter to make `woocommerce_form_field()` render valid Bootstrap markup.

Fixes #964

### DIFF
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -122,10 +122,13 @@ if ( ! function_exists( 'understrap_wc_form_field_args' ) ) {
 				$args['input_class'] = array( 'form-control' );
 				break;
 			case 'checkbox':
-				// Add a class to the form input's <label> tag.
-				$args['label_class'] = array( 'custom-control custom-checkbox' );
-				$args['input_class'] = array( 'custom-control-input' );
-				break;
+					$args['class'][] = 'form-group';
+					// Wrap the label in <span> tag.
+					$args['label'] = isset( $args['label'] ) ? '<span class="custom-control-label">' . $args['label'] . '<span>': '';
+					// Add a class to the form input's <label> tag.
+					$args['label_class'] = array( 'custom-control custom-checkbox' );
+					$args['input_class'] = array( 'custom-control-input' );
+					break;
 			case 'radio':
 				$args['label_class'] = array( 'custom-control custom-radio' );
 				$args['input_class'] = array( 'custom-control-input' );
--- a/inc/woocommerce.php
+++ b/inc/woocommerce.php
@@ -91,6 +91,7 @@ if ( ! function_exists( 'understrap_wc_form_field_args' ) ) {
 					'aria-hidden'      => 'true',
 				);
 				break;
+
 			/*
 			 * By default WooCommerce will populate a select with the country names - $args
 			 * defined for this specific input type targets only the country select element.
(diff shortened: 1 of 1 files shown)


IanDelMar:
@UnderstrapFramework please merge.

UnderstrapFramework:
Thanks

Thomas-A-Reinert:
10 merges within 3 minutes. It´s evident that no code-review has been done.
I´ll leave any interpretations open to the users of UnderStrap.

PR #1205, opened 2020-06-30
Added gulp deploy to upload dist-prod files to server

kingsolo50:
## Description

Added gulp ftp and gulp util to enable file upload to server.

## Motivation and Context

This will make the project more fully loaded able to cater for a project from start to finish.

## Types of changes

## Checklist:

### DIFF
--- a/gulpconfig.json
+++ b/gulpconfig.json
@@ -15,6 +15,9 @@
 		"composer": "./vendor",
 		"dev": "./src",
 		"dist": "./dist",
-		"distprod": "./dist-product"
+		"distprod": "./dist-product",
+		"host": "HOST FTP ADDRESS",
+		"user": "USERNAME",
+		"password": "PASSWORD"
 	}
 }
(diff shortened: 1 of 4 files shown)


danemorgan:
Hi @kingsolo50,

Thanks, this looks like it could be useful to me, but I think we have some work to do still.

What happens if someone fails to fill in the config file?

I wonder what the adoption rate on this will be. We want to be sure that anything we add to the code is going to be used by a good percentage of users. If only a few will be using it, it seems more appropriate to provide directions for them to add it.

I'm not sure if there is a way to conditionally include this, though presumably, we could comment it out, and then people who want to turn it on can uncomment it?

I think before we add anything that needs configuration as this does, we at least need to provide instructions in the readme to go along with it.

bacoords:
I'm going to cast my imaginary vote against merging this unless we see additional comments from other contributors. I find this to be a pretty specific use case and not very broadly applicable. I could definitely be wrong, but this seems like personal fork territory to me.

0dp:
I'm not a big fan of this.
Someone could naively distribute and expose ftp credentials by mistake.

Also I've never had the need to deploy my projects like this.
Just my two cents.

danemorgan:
Good enough for me, there are webhooks and other tools for this that seem safer.

PR #1250, opened 2021-01-29
RTL styles using gulp-rtlcss plugin

Milimas:
Gulp plugin that uses RTLCSS to convert LTR CSS to RTL instead of changing css manually.
feel free to correct my english

## Description

using gulp-rtlcss plugin instead of editing css or adding rtl.css file.

## Motivation and Context

this plugin make it easier to change sass files and distribute rtl css files just by saving sass file while running `$ glup watch` command

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have performed a self-review of my own code.
- [x] \(Optional) I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,25 @@ The front-page slider is widget driven. Simply add more than one widget to widge
 - That’s it.
 
 ## RTL styles?
-Add a new file to the themes root folder called rtl.css. Add all alignments to this file according to this description:
-https://codex.wordpress.org/Right_to_Left_Language_Support
+To change the style from LTR to RTL:
+- install [gulp-rtlcss](https://www.npmjs.com/package/gulp-rtlcss "gulp-rtlcss"):
+	- run: `$ npm install --save-dev gulp-rtlcss`
+- edit `gulpfile.js`:
+	- define `gulp-rtlcss` at the top of `gulpfile.js`:
+	
+		```javascript
+		var gulp = require('gulp');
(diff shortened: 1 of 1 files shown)


danemorgan:
Thanks @Milimas,

This PR does not actually include the RTL plugin in the gulpfile, instead it provides instructions for doing so in the readme. I'm in favor of approving this commit. It's in keeping with only building in things that a majority of users are going to need, but at the same time making them more accessible for those who might.

@bacoords, we might want to also adopt, and expand on this for our documentation project. As well as looking for other helpful plugins that might not be right for including, but that we can point people in the right directions.

bacoords:
@danemorgan Have we tested this method? I know we're just recommending it but not installing it (which makes sense to me) but I think we should test it out first.

I'd also be open to keeping both in the readme (and eventually documentation), the original/easy WordPress method and the more 'advanced' method as outlined here.

danemorgan:
@bacoords I agree. I have not merged it yet because I have not tested it to make sure it works.

Also, I want to see what you think of adding integrations.md, or scripts.md, or build.md file to collect these kinds of things in with a link from the readme.md.

bacoords:
As we're moving away from Gulp on the develop branch, I'm closing the PR. However, I HIGHLY encourage the original author to look into adding RTL support for the next release. We're using the same build process as Bootstrap and I know they have a lot of built in RTL stuff there.

PR #1273, opened 2021-06-11
Make the credits links https.

schnoggo:
## Motivation and Context
Just update links for a world where everything is https

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have checked that there aren't other open Pull Requests for the same update/change.
- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] \(Optional) My change requires a change to the documentation.
- [ ] \(Optional) I have updated the documentation accordingly.
- [ ] \(Optional) My change requires a change to the translations.
- [ ] \(Optional) I have updated the translations accordingly.
- [ ] `composer cs:check` has passed locally.
- [x] `composer lint:php` has passed locally.
- [x] I have read the **CONTRIBUTING** document.

Issue #930, opened 2019-03-04
Functions Refactoring 


misfist:
## Issue is about..
  - [ X  ] a Bug Report

## Environment:

  - **WordPress Version:** Version 5.1
  - **OS Version:** Linux [Ubuntu 18.04]
  - **Browser:** [ all  ]
  - **Language:** [ PHP ]
  - **Node.js & NPM Versions:** Node.js [v8.11.4 | NPM [V. 5.8.0]

After updating the Understrap from 0.8.7 to 0.8.8, I started getting an error in my custom child theme originating from the parent Understrap theme.

```
Uncaught Error: Call to undefined function understrap_categorized_blog() in /srv/www/metcouncil/wordpress/wp-content/themes/understrap/404.php:45
```

The issue seems to be due to changes made in the parent `functions.php`, which now uses `locate_template` to require files. The problem is that locate_template first looks in the child theme for the files, and my custom child theme has its own function files (sometimes) using the same paths. So, in this case, `understrap_categorized_blog()` can't be found because the function is in the parent `/inc/template-tags.php file`, but the locate_template only loads the child theme's `/inc/template-tags.php`, which doesn't contain that function.

As a work-around, I can rename the folder containing my child theme function files (say, `/includes` instead of `inc`). However, I'm not sure that the parent theme should be loading child function files rather than it's own. At the very least, this should be called out as a breaking change.

ref: https://codex.wordpress.org/Function_Reference/locate_template
understrap/understrap@9dacaa8

Duplicates: https://github.com/understrap/understrap-child/issues/214

## Steps to reproduce?
Create a understrap_child theme from the git repo. After activating the child theme, the widgets menu disappears under appearance. Among others.

## Who is the bug affecting?

All child themes that use the same file structure for custom functions (`/inc/setup.php`, `inc/extras.php`, etc)

## What is affected by this bug?

All parent theme settings and functionality if child theme contains same files.

## When does this occur?

Always

## Where on the platform does it happen?

Everywhere

## How do we replicate the issue?

After activating the child theme that is using understrap 0.8.8 as the parent. You can make the

## Are there any related Issues and/or PRs?

Maybe #845

Thomas-A-Reinert:
This is an UnderStrap-Child-Theme issue, so I´m not even going to read it at all.
Sorry, don´t get mad at me, but please post your question/issue in the https://github.com/understrap/understrap-child tracker. It just does not belong here.

misfist:
[...]

PR #805, opened 2018-09-12
Refactor functions.php

ylkyrg:
This dries up functions.php and makes it easier to read and edit.

PR #1279, opened 2021-06-15
Candidate for release v.0.9.6

### LINKED ISSUE #1278: Most recent update - messed up child theme.
Please delete anything that does not apply for you!

## Issue is about..
  - [ ] a Bug Report
  - [ ] I need help

  - **WordPress Version:** 5.7.2
  - **OS Version:** MacOS [Catalina - 10.15.7] | Android [not sure - s21 ] | iOS [iPad air gen 3 - 14.6]
  - **Browser:** [ all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari ]
  - **Language:** [ all | PHP | SQL | HTML | S/CSS | Javascript | jQuery | ES6 | JSON | XML]
  - **Node.js & NPM Versions:** Node.js [V X.XX] | NPM [V X.XX]

---

I updated to the newest version of Understrap - which I'm using a parent theme; however, when I updated the parent theme, it overwrote many theme files I had moved from the parent (with changes) changes. I've had to push a bunch of the function changes into my functions file to get them to work/display properly, which isn't how a child theme should work.

Things that are being affected is pagination, widgets, read more links.

 - the entire website. 

 - pagination
 - widgets
 - header styling 

- I didn't take an original backup, however, I've never had this issue before. 

## Expected behaviour
- for the parent theme not to override the child theme. 

#

danemorgan:
## Description

** Enhancements **
- Set credit links to use https protocol.

** Fixes **
- Use get_theme_file_path() to make include files child theme friendly.
	- Fixes breaking child theme overloading /inc/ folder files.

** Dependency Updates **
- Bump ws from 7.4.3 to 7.4.6
- Bump browserslist from 4.16.1 to 4.16.6
- Bump postcss from 8.2.9 to 8.2.10
- Bump hosted-git-info from 2.8.8 to 2.8.9
- Bump lodash from 4.17.20 to 4.17.21
- Bump ua-parser-js from 0.7.23 to 0.7.28

## Motivation and Context
Fix child theme file inclusion overloading, update dependencies, and make credit links https.

Fixes open issues
- https://github.com/understrap/understrap/issues/1278
- https://github.com/understrap/understrap/pull/1273

## Types of changes

## Checklist:

## Further comments
fixes #1278
closes #1273
closes #1269
closes #1267 
closes #1264 
closes #1263 
closes #1261

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -9,7 +9,7 @@
 defined( 'ABSPATH' ) || exit;
 
 // UnderStrap's includes directory.
-$understrap_inc_dir = get_template_directory() . '/inc';
+$understrap_inc_dir = 'inc';
 
 // Array of files to include.
 $understrap_includes = array(
@@ -40,5 +40,5 @@ if ( class_exists( 'Jetpack' ) ) {
 
 // Include files.
 foreach ( $understrap_includes as $file ) {
-	require_once $understrap_inc_dir . $file;
+	require_once get_theme_file_path( $understrap_inc_dir . $file );
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+## Release 0.9.6 June 15th 2021
+This is a bugfix and maintenance release.
+
+** Enhancements **
+
+- Set credit links to use https protocol.
+
+** Fixes **
+
+- Use get_theme_file_path() to make include files child theme friendly.
+
+	- Fixes breaking child theme overloading /inc/ folder files.
+
+** Dependency Updates **
+
+- Bump ws from 7.4.3 to 7.4.6
+- Bump browserslist from 4.16.1 to 4.16.6
+- Bump postcss from 8.2.9 to 8.2.10
+- Bump hosted-git-info from 2.8.8 to 2.8.9
+- Bump lodash from 4.17.20 to 4.17.21
+- Bump ua-parser-js from 0.7.23 to 0.7.28
+
+
 ## Release 0.9.5 June 10th 2021
 This is a maintenance release incorporating many commits including code formatting and dependency updates that have occured over the past two years. There are still many dependancy updates remaining and they will be addressed in new releases coming soon, but we want to get this release synced first.
 
--- a/inc/hooks.php
+++ b/inc/hooks.php
@@ -27,7 +27,7 @@ if ( ! function_exists( 'understrap_add_site_info' ) ) {
 
 		$site_info = sprintf(
 			'<a href="%1$s">%2$s</a><span class="sep"> | </span>%3$s(%4$s)',
-			esc_url( __( 'http://wordpress.org/', 'understrap' ) ),
+			esc_url( __( 'https://wordpress.org/', 'understrap' ) ),
 			sprintf(
 				/* translators: WordPress */
 				esc_html__( 'Proudly powered by %s', 'understrap' ),
@@ -37,7 +37,7 @@ if ( ! function_exists( 'understrap_add_site_info' ) ) {
 				/* translators: 1: Theme name, 2: Theme author */
 				esc_html__( 'Theme: %1$s by %2$s.', 'understrap' ),
 				$the_theme->get( 'Name' ),
-				'<a href="' . esc_url( __( 'http://understrap.com', 'understrap' ) ) . '">understrap.com</a>'
+				'<a href="' . esc_url( __( 'https://understrap.com', 'understrap' ) ) . '">understrap.com</a>'
 			),
 			sprintf( // WPCS: XSS ok.
 				/* translators: Theme version */
--- a/package-lock.json
+++ b/package-lock.json
@@ -7657,9 +7657,9 @@
       "dev": true
     },
     "nanoid": {
-      "version": "3.1.22",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.1.22.tgz",
-      "integrity": "sha512-/2ZUaJX2ANuLtTvqTlgqBQNJoQO398KyJgZloL0PZkC0dpysjncRUPsFe3DUPzz/y3h+u7C46np8RMuvF3jsSQ==",
+      "version": "3.1.23",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.1.23.tgz",
+      "integrity": "sha512-FiB0kzdP0FFVGDKlRLEQ1BgDzU87dy5NnzjeW9YZNt+/c3+q82EQDUwniSAUxp/F0gFNI1ZhKU1FqYsMuqZVnw==",
       "dev": true
     },
     "nanomatch": {
@@ -8357,9 +8357,9 @@
       "dev": true
     },
     "postcss": {
-      "version": "8.2.9",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.2.9.tgz",
-      "integrity": "sha512-b+TmuIL4jGtCHtoLi+G/PisuIl9avxs8IZMSmlABRwNz5RLUUACrC+ws81dcomz1nRezm5YPdXiMEzBEKgYn+Q==",
+      "version": "8.2.10",
+      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.2.10.tgz",
+      "integrity": "sha512-b/h7CPV7QEdrqIxtAf2j31U5ef05uBDuvoXv6L51Q4rcS1jdlXAVKJv+atCFdUXYl9dyTHGyoMzIepwowRJjFw==",
       "dev": true,
       "requires": {
         "colorette": "^1.2.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -6204,9 +6204,9 @@
       }
     },
     "hosted-git-info": {
-      "version": "2.8.8",
-      "resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.8.tgz",
-      "integrity": "sha512-f/wzC2QaWBs7t9IYqB4T3sR1xviIViXJRJTWBlx2Gf3g0Xi5vI7Yy4koXQ1c9OYDGHN9sBy1DQ2AB8fqZBWhUg==",
+      "version": "2.8.9",
+      "resolved": "https://registry.npmjs.org/hosted-git-info/-/hosted-git-info-2.8.9.tgz",
+      "integrity": "sha512-mxIDAb9Lsm6DoOJ7xH+5+X4y1LU/4Hi50L9C5sIswK3JzULS4bwk1FvjdBgvYR4bzT4tuUQiC15FE2f5HbLvYw==",
       "dev": true
     },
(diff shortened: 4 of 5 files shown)


bacoords (approved):
Looks great. Tested fine for me

functions.php
```diff
@@ -40,5 +40,5 @@
 
 // Include files.
 foreach ( $understrap_includes as $file ) {
-	require_once $understrap_inc_dir . $file;
+	require_once get_theme_file_path( $understrap_inc_dir . $file );
```
IanDelMar, line 43: Why would one want child themes to overload the parent theme's includes? It's the includes not the templates which are meant to be overloaded by child themes. This will most likely cause broken functionality due to missing functions. Each function is wrapped in `if ( ! function_exists( 'some_function_name' ) ) {}` and can easily be replaced with a function with the same name.

danemorgan: The last version of Understrap committed to the WordPress repository used get_template_part() to allow overloading /inc/ files and sat there for two years telling people to do exactly this.

When we updated to 0.9.5 we broke a number of sites that relied on this behavior, so we restored it with the appropriate function.

I'm open to understanding your concern about missing functions, but if someone copies an /inc/ file to their child folder, as I understand it now, WP will load the file from the child theme's /inc/ and not finding any function by that name already, load the function. If they delete the function from the child theme copy, it will still be in the parent theme folder and thus, undeclared, be included from there.

IanDelMar: @danemorgan I'm confused.

`get_theme_file_path()` allows child themes to overload the parent theme. It first looks in the stylesheet directory for the file and only if it does not exist in the stylesheet directory it looks in the parent theme's directory aka the template directory.

`get_template_directory()` on the other hand never looks in the stylesheet directory unless there is no child theme and the stylesheet directory is equal to the template directory. Note the word template. For a child theme the parent theme is the template that's why you have to put the template in the child theme's style.css. Here the Understrap child telling WordPress that Understrap is its _template_:
https://github.com/understrap/understrap-child/blob/0bb163325b2d24f15ebfbdb3fa144b866bdbeb12/style.css#L7

Correct me if I'm wrong but using `get_theme_file_path()` allows child themes to do this:
```php
// inc/template-tags.php

function child_slug_something() {
  return 'No more template tags from parent will break my site';
}

function child_slug_something_else() {
  return 'Now I have to copy each function to my child theme';
}

function child_slug_final_thought() {
  return 'Why am I allowed to do this in the first place?';
}
```

Some Understrap breaking changes history:
On Sep 12, 2018 @ylkyrg introduced `locate_template()` in #805. I think the change to giving child themes the permission to overload include files was unintentional and no one else noticed. Of course problems popped up, see, e.g., #930. As a consequence, on Jul 30, 2019 @ylkyrg made a PR to switch back to `get_template_directory` (#1020) - right after the release of 0.9.4 which was the last one for a long time. At that time Understrap started to die if I remember correctly. The new owner approved #1020 on Oct 29, 2019 but did not manage to merge the PR until Dec 13, 2019. But there was no release anyway. Meantime, people started to adjust and obviously were told to overload include files. You breaking the breaking change, of course, caused problems again, which led to the switch to `get_theme_file_path()` which has the same functionality as `locate_template()`.

bacoords: I think this is an interesting discussion point- how should child theme functions OVERRIDE the parent theme? 

In previous versions we did it with a `function_exists()` wrapper around the functions themselves. That allowed functions to be overridden on a per-function basis. Then in the last "release" update we included the option to override the entire PHP file as if it were a template file. I'm tempted to vote towards returning to the previous method. 

The biggest reason being that any new functionality included in those other `/inc/` files will not work, meaning we'll have to be even MORE liberal with our use of `function_exists()` throughout the rest of the theme in case we add a new function to an `/inc/` file and then try to use it in a template.

I understand the debate and the fact that we've now publicly supported various methods, but I guess if I were casting a vote, I would return to the era of overriding single functions, not being able to override an entire `inc/*.php` file full of functionality and potentially missing out on future features. 

That being said- to play devil's advocate: To build a child theme is to adapt a parent theme *at a certain point in time* and you may not want new functionality always popping up on a site you very specifically put together with a particular set of tools...

IanDelMar: I think missing out on a new feature is less of a problem but breaking sites with new features is. It will become tremendously cumbersome to always take care of function which might or might not be there because a child theme may have overloaded include files.

> That being said- to play devil's advocate: To build a child theme is to adapt a parent theme at a certain point in time and you may not want new functionality always popping up on a site you very specifically put together with a particular set of tools...

I disagree on this one. That's the whole point of using a child theme! The parent theme authors fix bugs, add new features and so on. If I wanted to get stuck with a theme and do this all by myself, I'd just use the parent, give it another slug/name and start modifying that theme. Then there is no need for a child theme. But I am aware that Understrap is kinf of a hybrid between parent and starter theme.

I really do think that it would be best to think about a work around. E.g. checking for the existence of a file in the child theme and then load the parent theme's file on top of that. Haven't thought much about this, but maybe there is something we can do. At least for those doing a first install of Understrap we should not provide the possibility of overloading the include files. We could even add the option to check the child theme files for conflicts with `get_template_directory()`, no? The long run outcome should really be people replacing functions and/or template files with child themes and nothing more. Even if this means more problems and support requests because of breaking sites now. Imagine a future situation where each single change to Understrap may break a child theme. That's nothing I want to work on, do you?
You're planning a major release. It's the perfect time to do it now before more and more people jump on the overload bandwagon.

danemorgan: I'm not insisting on the file override method. I restored it simply because when we pushed 0.9.5 to the WordPress repository we instantly started getting feedback that we had broken sites. I was not intending to advocate that position, only to explain the action and understand the ramifications more fully.

I actually thought of another potential problem, especially given the users involved. If someone overrides the file, then removes an entire function from the override, that function is gone, because the parent file isn't going to load, and that's got the potential to break the site too.

I'm fine with going back to stirctly using function overides, but we are going to have to manage a larger than insignificant WP repository users who are relying on that functionality.

Since our next version is going to be 1.0.0, I'd say we need to decide to do this now so it can go in there and not need another breaking release right away.

'm also in agreement with @IanDelMar's take that the purpose of a child theme is to allow continued updates and new features through the parent. If I want a feature-locked theme, I'm going to modify the core them, not use it as is with a child theme.

IanDelMar: Hi @danemorgan & @bacoords ! I had a look at a possible solution for this issue.

Basically, the following code first checks the child theme for the file. If it exists it loads the file. And then, no matter whether the file exist in the child theme or not all Understrap files get loaded. If a function has been defined in the child theme it won't be redefined because of all the `functions_exists()` checks and additionally because all Understrap files get loaded no function will be missing if a child theme overloaded a Understrap includes file in the past.

in functions.php
```diff
...

// Load Jetpack compatibility file if Jetpack is activiated.
if ( class_exists( 'Jetpack' ) ) {
	$understrap_includes[] = '/jetpack.php';
}

/*
 * In 2018 child themes were allowed to overload Understrap includes. This 
 * causes a lot of problems in the long run because Understrap is no longer in
 * control of its own includes.
 * Switching back to not allowing child theme to overload Understrap includes
 * introduces a breaking change for all child theme users. To mitigate the
 * effect we first load child theme files if they exist and load Understrap on
 * top. This ensures that each function of the child theme and each function
 * of Understrap gets loaded.
 */

// Include child theme files first.
+ if ( is_child_theme() ) {
+	foreach ( $understrap_includes as $file ) {
+		if ( file_exists( STYLESHEETPATH . $understrap_inc_dir . $file ) ) {
+			require_once get_stylesheet_directory() . $understrap_inc_dir . $file;
+		}
+	}
+}

// Load Understrap includes on top.
foreach ( $understrap_includes as $file ) {
-	require_once get_theme_file_path( $understrap_inc_dir . $file );
+	require_once get_template_directory() . $understrap_inc_dir . $file;
}
```

in inc/block-editor.php
```diff
...

if ( ! function_exists( 'understrap_generate_color_palette' ) ) {
	/**
	 * Checks for our JSON file of color values. If exists, creates a color palette array.
	 *
	 * @return array|bool
	 */
	function understrap_generate_color_palette() {
		$color_palette = array();

		// Grabs the autogenerated color palette that we're pulling from our compiled bootstrap stylesheets.
-		$color_palette_json = wp_remote_get( get_stylesheet_directory_uri() . '/inc/editor-color-palette.json' );
+		$color_palette_json = wp_remote_get( get_theme_file_uri( 'inc/editor-color-palette.json' ) );
		if ( ! is_wp_error( $color_palette_json ) ) {
			$color_palette_json = json_decode( $color_palette_json['body'], true );
			foreach ( $color_palette_json as $key => $value ) {
				$key             = ltrim( $key, '--' );
				$color_palette[] = array(
					'name'  => $key,
					'slug'  => $key,
					'color' => $value,
				);
			}
		}

		/**
		 * Filters the default bootstrap color palette so it can be overriden by child themes or plugins when we add theme support for editor-color-palette. This array can also be generated via gulp.
		 *
		 * @param array $color_palette An array of color options for the editor-color-palette setting.
		 */
		return apply_filters( 'understrap_theme_editor_color_palette', $color_palette );
	}
}

```

It worked fine in a first test on my local environment. Please feel free to test and tell me whether it worked for you as well.

If we implement this we could repeatedly tell people to check their files because we're going to deprecate the overloading sometime in the future. I think that's the long run goal. If one wants we could also add a database call and prohibit restrict existing users from overloading more files. I'm thinking about storing an option holding during the next update that holds all includes from the child theme and when loading the files only load those files from the child theme.

Assume a child theme comes with inc/extras.php and inc/hooks.php. On update we store the option understrap_overloads  with `array( 'extras.php' => true, 'hooks.php' => true )`. Then we can do the following in functions.php
```php
...

// Load Jetpack compatibility file if Jetpack is activiated.
if ( class_exists( 'Jetpack' ) ) {
	$understrap_includes[] = '/jetpack.php';
}

/*
 * In 2018 child themes were allowed to overload Understrap includes. This 
 * causes a lot of problems in the long run because Understrap is no longer in
 * control of its own includes.
 * Switching back to not allowing child theme to overload Understrap includes
 * introduces a breaking change for all child theme users. To mitigate the
 * effect we first load child theme files if they exist and load Understrap on
 * top. This ensures that each function of the child theme and each function
 * of Understrap gets loaded.
 */

// Include child theme files first.
 if ( is_child_theme() ) {
	$overloads = (array) get_option( 'understrap_overloads' );
	if ( ! empty( $overloads ) ) {
		foreach ( $understrap_includes as $file ) {
			if ( isset( $overloads[ $file ] ) ) {	
				if ( file_exists( STYLESHEETPATH . $understrap_inc_dir . $file ) ) {
					require_once get_stylesheet_directory() . $understrap_inc_dir . $file;
				}
			}
		}
	}
}
// Load Understrap includes on top.
foreach ( $understrap_includes as $file ) {
	require_once get_template_directory() . $understrap_inc_dir . $file;
}
```

I have no idea whether such a restriction that a child theme can not overload other files than it does at the time of the update would cause more problems than it solves.

PR #1309, opened 2021-07-09
Feature/gutenberg

bacoords:
## Description
This branch contains feature that support enhance the default experience for the "gutenberg" WordPress page editor.

### https://github.com/understrap/understrap/issues/1298 - default styles for blocks
A few blocks- blockquotes, tables, image block and cover blocks- are now fully supported AND make use of bootstrap's default styles when appropriate. For example, using the 'table' block will give you the same design as adding '.table' from Bootstrap.

### https://github.com/understrap/understrap/issues/1296 - enqueue basic styling into gutenberg
We're now actually adding styles to `custom-editor-styles.css` AND loading it into the Gutenberg page editor. This means that we're loading fonts, typography, spacing, sizing, colors, and some custom block styles into the back end. This stylesheet is built using the same gulp process as before, except it now is importing a few tools from Bootstrap to render the necessary matching CSS.

### https://github.com/understrap/understrap/issues/1295 - bootstrap theme colors in gutenberg.
Instead of letting gutenberg provide a random color palette as the default picker options, we've added a new task to the gulp workflow that parses out the color variables from the final stylesheet and outputs them as `inc/editor-color-palette.json`. Then we use the `add_theme_support( 'editor-color-palette' )` function to parse that json file and send those colors into Gutenberg. So if you were to change the SCSS variable `$purple` and compile the stylesheets, that shade of purple will be a default color option on the back end.

### https://github.com/understrap/understrap/issues/1294 - Alignwide/alignfull
This PR includes support for the `alignwide` and `alignfull` alignment options for images and cover blocks. In order to allow this functionality, we run a check on the current template and customizer settings in `inc/extras.php` that determines the presence of a sidebar and then adds a body class. The alignments will only stretch out of their containers if the `understrap-no-sidebar` class is present, otherwise, they'll simply remain regular width.

## Types of changes
- Breaking change (fix or feature that would cause existing functionality to change)
I'm going to call this a 'breaking change' because we're essentially changing how the backend editor works (it now loads in a lot more styles) and how some blocks display on the front end. It shouldn't 'break' any sites, but it will provide a different backend experience and it will include a few minor front-end styles that weren't there before.

## Checklist:

## Further comments
I would love it if others wanted to pull this down and test it out, and potentially submit any other ideas. The goal is to have the next major version (1.0.0) be a full support of Gutenberg. I'm going to work on a similar branch for the child theme now as well, so would love any thoughts/feedback on the changes here.


IanDelMar:
Hi @bacoords! I had a look at the color palette generation stuff today. Am I right that it is required to import Bootstrap's _root.scss at least in one of the main scss files, custom-editor-style.scss or theme.scss to get the color palette to work?

Also, are you planning to introduce breaking changes to the next major release (do we finally have semantic versioning)? Because there a couple of other breaking change things that could be included then.

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+		$color_palette = array();
+
+		// Grabs the autogenerated color palette that we're pulling from our compiled bootstrap stylesheets.
+		$color_palette_json = wp_remote_get( get_stylesheet_directory_uri() . '/inc/editor-color-palette.json' );
+		if ( ! is_wp_error( $color_palette_json ) ) {
+			$color_palette_json = json_decode( $color_palette_json['body'], true );
+			foreach ( $color_palette_json as $key => $value ) {
+				$key             = ltrim( $key, '--' );
+				$color_palette[] = array(
+					'name'  => $key,
```
IanDelMar, line 49: The color name is supposed to be a translatable string. Non-English speaking screen reader users won't know which color it is unless you provide a translation. This is a no go in terms of accessibility.

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+			add_theme_support( 'editor-color-palette', $color_palette );
+		}
+	}
+}
+
+if ( ! function_exists( 'understrap_generate_color_palette' ) ) {
+	/**
+	 * Checks for our JSON file of color values. If exists, creates a color palette array.
+	 *
+	 * @return array|bool
```
IanDelMar, line 37: When does the function have a boolean return type?

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	function understrap_generate_color_palette() {
```
IanDelMar, line 39: I'd go for `understrap_get_color_palette()`. However, that's just a matter of taste.

inc/block-editor.php
bacoords: If they've chosen to override with a filter it could theoretically return false to cancel it out - although I guess if they want to override it they should return an empty array....

inc/block-editor.php
bacoords: Interesting. In that case - I see no reason not to skip the loop function and filter overall. I can hard code the entire thing with the default list of colors in bootstrap (blue, red, etc) while still pulling the hex 'value' from the json file. When I generate the list in gulpconfig.js I specify the colors there, so the precedent exists.

gulpconfig.json
```diff
@@ -16,5 +16,20 @@
 		"dev": "./src",
 		"dist": "./dist",
 		"distprod": "./dist-product"
-	}
+	},
+	"colors": [
```
IanDelMar, line 20: What about primary, secondary, light, dark? I'm not sure about the alert colors. The CSS is compiled for all the colors in the colors and the theme colors map.

inc/block-editor.php
IanDelMar: You're not type casting the return value. It can be whatever the user wants it to be ;-)

sass/theme/_colors.scss
```diff
@@ -0,0 +1,22 @@
...
+
+	.has-#{$name}-color,
+	.has-#{$name}-color:visited {
+		color: $color;
+	}
+	.has-#{$name}-background-color {
+		background-color: $color;
+		border-color: $color;
+	}
+}
```
IanDelMar, line 22:
```suggestion
@each $name, $color in map-merge($colors, $theme-colors) {

	.has-#{$name}-color,
	.has-#{$name}-color:visited {
		color: $color;
	}
	.has-#{$name}-background-color {
		background-color: $color;
		border-color: $color;
	}
}

```

sass/understrap/understrap.scss
```diff
@@ -42,18 +42,18 @@
 .alignleft {
 	display: inline;
 	float: left;
-	margin-right: 1.5em;
+	margin-right: $spacer;
```
IanDelMar, line 45: By default $spacer is 1rem which probably is not equal to 1.5em. This changes frontend styles which are not necessarily related to Gutenberg.

sass/understrap/understrap.scss
```diff
@@ -42,18 +42,18 @@
...
 }
 
 .alignright {
 	display: inline;
 	float: right;
-	margin-left: 1.5em;
+	margin-left: $spacer;
```
IanDelMar, line 51: Same as above.

gulpfile.js
```diff
@@ -106,6 +108,31 @@ gulp.task( 'minifycss', function() {
 		.pipe( gulp.dest( paths.css ) );
 } );
 
+
+/**
+* Minifies css files.
+*
+* Run: gulp minifycss
+*/
```
IanDelMar, line 116:
```suggestion
/**
 * Generates inc/editor-color-palette.json based on the colors defined
 * in gulpconfig.json.
 * 
 * Run: gulp gen-palette
 */
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+
+add_action( 'after_setup_theme', 'understrap_block_editor_setup' );
+
+if ( ! function_exists( 'understrap_block_editor_setup' ) ) {
+
+	/**
+	 * Sets up our default theme support for the WordPress block editor.
+	 *
```
IanDelMar, line 15:
```suggestion
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Block editor (gutenberg) specific functionality
```
IanDelMar, line 3:
```suggestion
 * Block editor (Gutenberg) specific functionality
```

sass/theme/_blocks.scss
```diff
@@ -0,0 +1,75 @@
+  // Tables
```
IanDelMar, line 1:
```suggestion
// Tables
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @package UnderStrap
+ */
+
+
```
IanDelMar, line 8:
```suggestion
```

sass/theme.scss
```diff
@@ -8,4 +8,6 @@
 
 // Any additional imported files //
 @import "theme/theme";            // <-------- That's where you can add your own design. Thats your part!
+@import "theme/colors";           // <-------- This creates the necessary bootstrap color classes.
```
IanDelMar, line 11:
```suggestion
@import "theme/colors";           // <-------- This creates the necessary Bootstrap color classes.
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+}
+
+if ( ! function_exists( 'understrap_generate_color_palette' ) ) {
+	/**
+	 * Checks for our JSON file of color values. If exists, creates a color palette array.
+	 *
+	 * @return array|bool
+	 */
```
IanDelMar, line 38:
```suggestion
	/**
	 * Retrieves an array of the theme's custom color palette for the block
	 * editor.
	 *
	 * @return array
	 */
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+					'color' => $value,
+				);
+			}
+		}
+
+		/**
+		 * Filters the default bootstrap color palette so it can be overriden by child themes or plugins when we add theme support for editor-color-palette. This array can also be generated via gulp.
+		 *
+		 * @param array $color_palette An array of color options for the editor-color-palette setting.
+		 */
```
IanDelMar, line 60:
```suggestion
		/**
		 * Filters the theme's custom color palette for the block editor.
		 *
		 * @param array[] $color_palette The custom color_palette.
		 */
```

howarddc:
@IanDelMar Thank you for all your contributions - we really appreciate it! :-)

gulpfile.js
```diff
@@ -121,7 +148,7 @@ gulp.task( 'cleancss', function() {
  * Run: gulp styles
  */
 gulp.task( 'styles', function( callback ) {
-	gulp.series( 'sass', 'minifycss' )( callback );
+	gulp.series( 'sass', 'minifycss', 'gen-palette' )( callback );
```
IanDelMar, line 151:
```suggestion
	gulp.series( 'sass', gulp.parallel( 'minifycss', 'gen-palette' ) )( callback );
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+if ( ! function_exists( 'understrap_generate_color_palette' ) ) {
+	/**
+	 * Checks for our JSON file of color values. If exists, creates a color palette array.
+	 *
+	 * @return array|bool
+	 */
+	function understrap_generate_color_palette() {
+		$color_palette = array();
+
+		// Grabs the autogenerated color palette that we're pulling from our compiled bootstrap stylesheets.
```
IanDelMar, line 42:
```suggestion
		/**
		 * Grabs the autogenerated color palette that we're pulling from our
		 * compiled Bootstrap stylesheets.
		 */
```

inc/block-editor.php
```diff
@@ -0,0 +1,63 @@
...
+		$color_palette_json = wp_remote_get( get_stylesheet_directory_uri() . '/inc/editor-color-palette.json' );
```
IanDelMar, line 43:
```suggestion
		$color_palette_json = wp_remote_get( get_theme_file_uri() . '/inc/editor-color-palette.json' );
```

A child theme not shipping its own /inc/editor-color-palette.json will throw an error when using `get_stylesheet_directory_uri()`.

danemorgan:
@IanDelMar We are trying to use semantic versioning going forward.

The plan is for the next version to be 1.0.0, and breaking changes are possible with that. maybe we are judicious and try not to overwhelm though. 😂

inc/block-editor.php
bacoords: This is a good suggestion.

PR #1313, opened 2021-07-20
[WIP] Update the search form

IanDelMar:
## Description

This PR update the search form in the following ways:
- It adds the $args parameter/the aria-label attribute which was added in WP 5.2.0
- It removes the IDs from the form tag itself and from the submit input field. They are unnecessary and would have had to be changed anyway (see motivation).
- It makes the ID for the search input unique (see motivation).  
- It adds context to the message ids and updates translations accordingly.
- It adds those classes to the elements which WordPress uses for the search form.

## Motivation and Context

Getting in line with the WordPress default search form again. Also, there might by more than 1 search form per page which then would require unique IDs for the different forms.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

It probably increases the required WP version from 4.7 to 5.0 as `wp_unique_id()` was introduced in WP 5.0.3 (see https://github.com/understrap/understrap/issues/1183#issuecomment-636988539). It also breaks CSS and JS which relies on the IDs that have been removed or changed. I think neither the WP version increase nor the removal/change of the classes will affect many people.

## Checklist:

- [ ] (Optional) My change requires a change to the documentation.
- [ ] (Optional) I have updated the documentation accordingly.
- [x] (Optional) My change requires a change to the translations.
- [x] (Optional) I have updated the translations accordingly.

## Further comments

Adding the following function and using it instead of `wp_unique_id()` would not increase the required WP version:
```php
function understrap_unique_id( $prefix = '' ) {
  global $wp_version;
  if ( version_compare( $wp_version, '5.0.3' ) >= 0) {
    wp_unique_id( $prefix ):
  }
  static $id_counter = 0;
  return $prefix . (string) ++$id_counter;
}
```

I'd suggest to also drop the classes `.field` and `.submit`. But this would add another layer of possible breaking change. Again, I think that this won't have much of an impact on users.


Thomas-A-Reinert (commented):
Hey @IanDelMar to make the search function fully accessible, you´d need to follow these links if I understand that code correctly..

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role
- http://web-accessibility.carnegiemuseums.org/code/search/
- https://stackoverflow.com/a/43860344
- https://www.a11ymatters.com/pattern/accessible-search/
- https://www.bmas.de/DE/Startseite/start.html – pretty good (yet not perfect) implementation example (main nav "Suche" Widget), following WCAG 2.1, Section 508, BITV 2.0 (actually 2.1) Standards
- https://www.bmas.de/SiteGlobals/Forms/Suche/Servicesuche_Formular.html?templateQueryString=understrap - Actually pretty bad example, where role="search" appears twice on a page.


**What the actual implementation lacks:**
- there is no headline or alike (for sr/for (visible) users using the site without css.
- Yes, aria roles help a lot, but they should be used sparingly and only when there is no native html tag matching that role.
If that thing going to be re-used on a site (in main navigation and in a widget), only ONE should be aria-flagged as role="search". And please keep the focus on the element, that´s persistant on EVERY page (the one in the main navigation).

The last link may be the most interesting to really _understand_ the whole problematic situation for an impaired user: https://www.a11ymatters.com/pattern/accessible-search/ 

- Handling form errors
- Always include a clear focus style
- If the input is required, indicate that
- Building the search component

Thomas-A-Reinert:
Sidenote: I can´t approve/disapprove a merge/pull request. Or add a review.
image

IanDelMar:
@Thomas-A-Reinert thanks for your feedback. I'll have a look at _all_ the links.

> In a WP World that could happen – but should definitely be avoided – by adding a search widget to the main navigation and the sidebar. Or a searchpage, that re-uses the same function. To be clear: There should only be ONE role="search" on a page.

Who's going to tell the user who is free to add as many search forms wherever they want?

> there is no headline or alike (for sr/for (visible) users using the site without css.

According to https://www.a11ymatters.com/pattern/accessible-search/ and https://www.w3.org/WAI/tutorials/forms/labels/ it's ok to hide the label as the context is visually clear.

> Yes, aria roles help a lot, but they should be used sparingly and only when there is no native html tag matching that role.
If that thing going to be re-used on a site (in main navigation and in a widget), only ONE should be aria-flagged as role="search". And please keep the focus on the element, that´s persistant on EVERY page (the one in the main navigation).

There is a native html tag matching the search role??? We don't know which search form is persistent on every page. Again, that's up to the user. The only thing we can guess is: if a search form is in the main nav it's probably persistent.

searchform.php
```diff
@@ -7,16 +7,24 @@
...
-	<div class="input-group">
-		<input class="field form-control" id="s" name="s" type="text"
-			placeholder="<?php esc_attr_e( 'Search &hellip;', 'understrap' ); ?>" value="<?php the_search_query(); ?>">
-		<span class="input-group-append">
-			<input class="submit btn btn-primary" id="searchsubmit" name="submit" type="submit"
-			value="<?php esc_attr_e( 'Search', 'understrap' ); ?>">
-		</span>
+<form class="search-form" method="get" action="<?php echo esc_url( home_url( '/' ) ); ?>" <?php echo $aria_label; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Escaped above. ?>>
+	<div role="search">
+		<label class="sr-only" for="<?php echo $uid; ?>"><?php echo esc_html_x( 'Search for:', 'label', 'understrap' ); ?></label>
```
bacoords, line 21: Should we escape `$uid` here and below?

IanDelMar: `wp_unique_id()` is considered save as it only adds an integer after the prefix and in this case our prefix is safe too.

```
function wp_unique_id( $prefix = '' ) {
    static $id_counter = 0;
    return $prefix . (string) ++$id_counter;
}
```
Code taken from  https://developer.wordpress.org/reference/functions/wp_unique_id/

danemorgan:
> Sidenote: I can´t approve/disapprove a merge/pull request. Or add a review.
> image

I'll look at that. You should be able to review, so I misconfigured something.

IanDelMar:
I reverted adding the `role="search"`on an extra div. The argument made at https://www.a11ymatters.com/pattern/accessible-search/ is outdated. The prohibition of role="search" on the form element was lifted in 2016. 

I didn't find any resource telling to use _one_ role="search" only. Yeah, landmark roles should be used sparingly. But even Mozilla  gives an example of using role="search" twice on a page.
```html
<header>
  <form id="site-search-top" role="search" aria-label="Sitewide">
    <!-- search input -->
  </form>
</header>

...

<footer>
  <form id="site-search-bottom" role="search" aria-label="Sitewide">
    <!-- search input -->
  </form>
</footer>
```
One should use the same label for the forms then. In our case this means to also use `aria-label="Sitewide"` because the default search is sitewide. If we choose "Pages" or "All Pages" or "Site" we could even use the core's translation for it. I don't know whether that sounds weird to a native though. Like so
```php
$aria_label = 'aria-label="' . esc_html_x( 'Sitewide', 'search', 'understrap' ) . '"';
if ( isset( $args['aria_label'] ) && ! empty( $args['aria_label'] ) ) {
	$aria_label = 'aria-label="' . esc_attr( $args['aria_label'] ) . '"';
}
```

0dp:
> > Sidenote: I can´t approve/disapprove a merge/pull request. Or add a review.
> > image
> 
> I'll look at that. You should be able to review, so I misconfigured something.

I think it's a nice feature bit might be annoying that all requests has to get a review.

But it could be as simple as an ok

danemorgan:
Yes. The primary purpose for requiring all of them to have a review is that at a minimum someone has to tick the "approve" box first. :)

bacoords (approved):

bacoords (dismissed):

bacoords:
I'd like to merge this in but there's too many conflicts with the language files that I don't feel comfortable handling. I'll leave this open until someone can pull it closer in line with the `develop` branch

IanDelMar:
I solved the conflicts but let me tell you that I didn't feel comfortable neither. Consider this my last act so that there is no open PR from me any longer.

bacoords (approved):

bacoords:
@IanDelMar Your work on this project has been greatly appreciated.

PR #1317, opened 2021-07-22
Add dependabot config file

IanDelMar:
## Description

This PR adds a config file for the GitHub-native Dependabot.

## Motivation and Context

Let Dependabot check for outdated dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,9 @@ updates:
   # Maintain dependencies for npm
   - package-ecosystem: "npm"
     directory: "/"
+    ignore:
+      - dependency-name: "bootstrap"
+        versions: "5.x"
     labels:
     - "dependencies"
     open-pull-requests-limit: 10
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,36 @@
+
+# Set up for npm and Composer
+
+version: 2
+updates:
+
+  # Maintain dependencies for npm
+  - package-ecosystem: "npm"
+    directory: "/"
+    labels:
+    - "dependencies"
+    open-pull-requests-limit: 10
+    rebase-strategy: auto
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "06:00"
+    target-branch: "develop"
+    versioning-strategy: increase
+
+  # Maintain dependencies for Composer
+  - package-ecosystem: "composer"
+    allow:
+    - dependency-type: direct
+    - dependency-type: indirect
+    directory: "/"
+    labels:
+    - "dependencies"
+    open-pull-requests-limit: 10
+    rebase-strategy: auto
+    schedule:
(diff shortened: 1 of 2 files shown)


danemorgan (approved):

IanDelMar:
Side note: Maybe the dependabot file needs to be placed in the main branch before dependabot starts making PRs to the develop branch.

danemorgan:
Thanks @IanDelMar 

Quick question on that. Should develop *not* have it also, or it just needs to be in both?

Also, what are the ramifications of dependabot issuing PRs against develop, and then merging them into main from there?

Is it just the delay that would occur and we need to get the dependabot commits into the main faster?

danemorgan:
It looks like this PR was checked out from develop, so merging to main would carry everything in develop right now.

IanDelMar:
> Quick question on that. Should develop not have it also, or it just needs to be in both?

Develop will have it also. While the develop branch might be ahead of the master it will never be the other way around. My point was that I think dependabot looks in the main branch for the config file and won't do anything as long it doesn't find the file there. So if we want dependabot to start now we have to put the config file there as well. 

> Also, what are the ramifications of dependabot issuing PRs against develop, and then merging them into main from there?

The config file tells dependabot to make PRs against the develop branch. 
https://github.com/understrap/understrap/blob/989c0ffb53120c217f14fadbc1e13fec4d44e4bb/.github/dependabot.yml#L21

> Is it just the delay that would occur and we need to get the dependabot commits into the main faster?

There's no need for this if you have releases regularly. If not then I would recommend to update the dependencies in both branches. So people using the install from the WP repository don't end up having outdated packages with potential security issues (I don't know if people use the node stuff on live sites) branch One can configure dependabot to make PR against both branches. Maybe that would be the best.

IanDelMar:
> It looks like this PR was checked out from develop, so merging to main would carry everything in develop right now.

Yes, because I read that one should use the develop branch for PRs. I'll make a PR another PR for merging the file in the main branch.

IanDelMar:
Should we add the template chooser to the main branch as well? Otherwise it cannot be used until develop gets merged into main.

danemorgan:
I think it would be nice to deploy that early too, yes.

The develop for PRs is the general rule. You're doing special stuff. Any time you think something needs to go into main early, go ahead and think of it as a hotfix. We can still merge it into develop also, but it will let us get it into main quickly.

danemorgan:
> So people using the install from the WP repository don't end up having outdated packages with potential security issues (I don't know if people use the node stuff on live sites)

We actually have to strip out all of the packages and build files before we submit them to the WordPress repository. People using the theme from the repository have no access to this stuff. I suspect most of them just use it as a normal parent theme and make a child theme. Hence the issue with them being behind the curve on the whole file overloading thing.

IanDelMar:
Yes, maybe it's sufficient to have the latest packages at the develop branch.

PR #1380, opened 2021-08-18
[WIP] Move build process away from gulp

bacoords:
## Description
This moves us away from gulp and instead uses npm based tasks, so instead of `gulp watch` we have `npm run watch` and so on. For sass, we're using PostCSS. Instead of Webpack or Laravel Mix, I went with Rollup as that is what the Bootstrap team has used on v4 and v5. Many of the commands are based on Bootstrap, but simplified. 

This also means that we're now importing all of the bootstrap modules directly into `custom-javascript.js`, which means end users can remove unneeded code much easier or import new libraries without messing with the gulp files.

This PR does not include the following gulp tasks
- `gulp clean / gulp dist  / etc` - still looking into these
- `gulp imagemin` - does anyone use this?
- `gulp copy-assets` - we can perhaps replace this, but I'm not convinced it's needed. I see no reason not to reference node_modules in the build process?
- `gulp gen-palette` - still working on an alternative for this

## Motivation and Context

https://github.com/understrap/understrap/issues/1379

## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x] I have checked that there aren't other open Pull Requests for the same update/change.
- [ x] My code follows the code style of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have read the **CONTRIBUTING** document.


src/js/custom-javascript.js
```diff
@@ -1 +1,36 @@
+/**
+ * --------------------------------------------------------------------------
+ * Understrap (v1.0.0): index.js
+ * Licensed under GPL (http://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html)
+ * --------------------------------------------------------------------------
+ */
+
+import Alert from '../../node_modules/bootstrap/js/src/alert'
```
0dp, line 8: use `https://github.com/rollup/plugins/tree/master/packages/node-resolve` so that you don't need the `.../.../node_modules/` part

package.json
```diff
@@ -31,10 +37,15 @@
...
-  "dependencies": {},
+  "dependencies": {
+    "terser": "^5.7.1"
+  },
   "devDependencies": {
     "@babel/core": "^7.14.8",
     "@babel/preset-env": "^7.14.9",
+    "@rollup/plugin-babel": "^5.3.0",
+    "@rollup/plugin-commonjs": "^20.0.0",
+    "@rollup/plugin-node-resolve": "^13.0.4",
```
0dp, line 48: You got it here, so use it in custom-javascript.js

src/js/custom-javascript.js
```diff
@@ -1 +1,36 @@
+/**
```
0dp, line 1: don't call it custom-javascript.js. just call it bootstrap.js and leave custom-javasctipt.js for custom javascript

package.json
```diff
@@ -4,12 +4,18 @@
   "description": "WordPress Theme framework",
```
0dp, line 4: These scripts are more or less c+p from BS right?

package.json
```diff
@@ -4,12 +4,18 @@
...
-    "css-postcss": "postcss --config postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
-    "watch": "npm-run-all --parallel watch-*",
-    "watch-css": "nodemon --watch sass/ --ext scss --exec \"npm-run-all css\""
+    "css-postcss": "postcss --config build/postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
+    "js": "npm-run-all js-compile js-minify",
+    "js-compile": "rollup --config build/rollup.config.js --sourcemap",
+    "js-minify": "terser --compress typeofs=false --mangle --comments \"/^!/\" --source-map \"content=js/theme.js.map,includeSources,url=theme.min.js.map\" --output js/theme.min.js js/theme.js",
+    "watch": "npm-run-all --parallel watch-run-*",
+	"watch-bs": "npm-run-all --parallel bs watch-run-*",
+    "watch-run-css": "nodemon --watch sass/ --ext scss --exec \"npm-run-all css\"",
```
0dp, line 17: nodemon is not a dev dependency in package.json and do we need nodemon at all?
This is per see not a node.js app

package.json
```diff
@@ -4,12 +4,18 @@
...
     "css": "npm-run-all css-compile css-postcss css-minify",
     "css-compile": "sass --style expanded --source-map --embed-sources --no-error-css sass/theme.scss:css/theme.css sass/custom-editor-style.scss:css/custom-editor-style.css",
     "css-minify": "cleancss -O1 --format breakWith=lf --with-rebase --source-map --source-map-inline-sources --output css/ --batch --batch-suffix \".min\" \"css/*.css\" \"!css/*.min.css\" \"!css/*rtl*.css\"",
-    "css-postcss": "postcss --config postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
-    "watch": "npm-run-all --parallel watch-*",
-    "watch-css": "nodemon --watch sass/ --ext scss --exec \"npm-run-all css\""
+    "css-postcss": "postcss --config build/postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
+    "js": "npm-run-all js-compile js-minify",
+    "js-compile": "rollup --config build/rollup.config.js --sourcemap",
+    "js-minify": "terser --compress typeofs=false --mangle --comments \"/^!/\" --source-map \"content=js/theme.js.map,includeSources,url=theme.min.js.map\" --output js/theme.min.js js/theme.js",
```
0dp, line 14: should probably add a terser config file or options to customize terser options.
ie. CI/CD might only be able to run on a single thread where i might want to run it on 4

package.json
```diff
@@ -4,12 +4,18 @@
...
   "main": "index.js",
   "scripts": {
+	"bs": "browser-sync start --config build/browser-sync.config.js",
     "css": "npm-run-all css-compile css-postcss css-minify",
     "css-compile": "sass --style expanded --source-map --embed-sources --no-error-css sass/theme.scss:css/theme.css sass/custom-editor-style.scss:css/custom-editor-style.css",
     "css-minify": "cleancss -O1 --format breakWith=lf --with-rebase --source-map --source-map-inline-sources --output css/ --batch --batch-suffix \".min\" \"css/*.css\" \"!css/*.min.css\" \"!css/*rtl*.css\"",
-    "css-postcss": "postcss --config postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
-    "watch": "npm-run-all --parallel watch-*",
-    "watch-css": "nodemon --watch sass/ --ext scss --exec \"npm-run-all css\""
+    "css-postcss": "postcss --config build/postcss.config.js --replace \"css/*.css\" \"!css/*.rtl*.css\" \"!css/*.min.css\"",
```
0dp, line 11: we need a `postcss.config.js`

package.json
bacoords: Exactly, we're thinking it makes the most sense to follow what they do in BS4/5. I just trimmed a lot of them down and also tried to bring the naming conventions closer to what we had before ('gulp watch' becomes 'npm run watch')

package.json
bacoords: It's in the `build/` folder. This was another convention from BS that I thought made sense, putting as much of the config files in a separate folder. When we submit to wp.org repo, we have to remove everything build related, so having them in one folder seems helpful.

package.json
bacoords: I'm using nodemon to watch our asset files for changes and re-run all of the compile/minify/etc scripts. Open to alternatives though

package.json
0dp: But the term build in my eyes have to do with the final build not general build. but that is just my POV.
I have no issue with having config files in the root. But if its a convention thing sure :)

package.json
0dp: rollup.js' `watch` command maybe `-w, --watch                 Watch files in bundle and rebuild on changes`

src/js/custom-javascript.js
bacoords: 👍

src/js/custom-javascript.js
bacoords: Good idea- I see I also wasn't bundling the skip-link-focus-fix.js file, so I fixed that. But now you can just modify `src/js/bootstrap.js` to remove any unnecessary bootstrap modules.

package.json
bacoords: Oh that's cool - but I'd need something for the scss?

0dp: rollup-plugin-scss

Otherwise it works just like webpack where you include your scss in index.js and extract it

package.json
bacoords: Yeah I keep going back and forth on this. I don't really like having a new directory but I also would love to have all the dev-type stuff collected together so we can more easily prepare for wp.org...

package.json
bacoords: Can you send example? I didn't see one in Bootstrap so I'm not sure what it would look lik

package.json
bacoords: In that case I'll stick with the default way Bootstrap handles it, using nodemon. 

I'm wondering if it's worth moving the sass folders though, it's always bothered me and people we teach that there's `/sass/` and `/src/sass/`...

package.json
0dp: You see how they are calling terser and a series of flags
```
terser --compress typeofs=false --mangle --comments \"/^!/\" --source-map \"content=js/theme.js.map,includeSources,url=theme.min.js.map\" --output js/theme.min.js js/theme.js
```
That is not a very sustainable way of doing it for instance if I want to toggle `parallel` based on a specific scenario dev/prod build.

Here is an example from a project i got with webpack
```
            minimizer: [
                (compiler) => {
                    const TerserPlugin = require('terser-webpack-plugin');
                    new TerserPlugin({
                        parallel: true,
                        test: /\.js($|\?)/i,
                        // sourceMap: IS_DEV,
                        terserOptions: {
                            compress: {
                                booleans: IS_PROD,
                                conditionals: IS_PROD,
                                drop_console: IS_PROD,
                                drop_debugger: IS_PROD,
                                if_return: IS_PROD,
                                join_vars: IS_PROD,
                                keep_classnames: IS_DEV,
                                keep_fnames: IS_DEV,
                                reduce_vars: IS_PROD,
                                sequences: IS_PROD,
                                warnings: IS_DEV,
                                ecma: 5,

                            },
                            output: {
                                comments: IS_DEV,
                            },
                        },
                    }).apply(compiler);
                },
                new OptimizeCSSAssetsPlugin({}),
            ],
```

You can do something similar with `rollup-plugin-terser`

package.json
0dp: all scss and js folders should be in /src

0dp: > In that case I'll stick with the default way Bootstrap handles it, using nodemon.
> 
> I'm wondering if it's worth moving the sass folders though, it's always bothered me and people we teach that there's `/sass/` and `/src/sass/`...

just remember to add nodemon as a dev dep then.

Of course do you, I just think it's a cleaner approach to have the jobs the tasks run defined in a config file for that purpose ie. rollup.config.js

package.json
0dp: do we actually need to clean the dev stuff out for wp.org?
If so then put that folder in /src as well

bacoords:
Update on this PR- 

I've replaced the `gulp gen-palette` command with a [custom postcss plugin](https://github.com/bacoords/postcss-understrap-palette-generator). This plugin basically reads your color variables from your compiled stylesheet and creates the json file needed for the theme to populate the Gutenberg color palette. I hope to make this plugin more full-featured someday so that it's useful outside of Understrap.

As of now, I'm seeing no reason to recreate the `imagemin` task. We don't really even bundle images in the theme, so adding a task for something we're not doing by default seems unnecessary.

I'm also on the fence about all of the `dist` and `clean` type tasks. The `copy-assets` task was handy but in my mind I'm not sure why we need all of the assets duplicated in the theme anyway. If you're running the `npm` scripts, you need your `node_modules` folder populated, so why add a bunch of files to the theme? Or we just add them manually- does it need a task But I could be persuaded to quickly throw it together using [copyfiles](https://preview.npmjs.com/package/copyfiles). Either way I'm still reviewing this.

The last thing I plan to do here is combine `/sass/` and `/src/sass/` so that all developer assets are in one place, and we have one less folder on the site. This is a constant point of confusion for newer developers

package.json
bacoords: That's a good point- maybe /src makes more sense and gives us one less directory. Or we just put it back in the root since that's pretty standard and we have to delete a lot of stuff from root anyway. 

We don't have to remove asset files (like scss) but we have to remove any config files as shown here https://make.wordpress.org/themes/handbook/review/required/#9-files

package.json
bacoords: 👍 New issue to track this here: https://github.com/understrap/understrap/issues/1400

PR #1435, opened 2021-09-18
Bring back themejs folder watch

gettonet:
Earlier version had gulp task watch themejs folder for js files, now that functionality is gone.

This feature is very useful because it allowed putting any library in a separate .js file in that folder, and keep custom-javascript.js for custom scripts.


gettonet:
I made a mistake, the wildcard I added is not recognized in node.js, I apologize.

PR #1661, opened 2022-03-13
WIP: Optimize contact form 7 integration

bacoords:
The Contact Form 7 integration adds a lot of CSS that is most likely not needed for most users. I'd love to see it either spun out as an optional CSS file that loads only if the plugin loads, or else neatened up as much as possible. 

Here's what's been done so far 
- [x] Uses the button mixin instead of extending the button classes
- [x] Removes the form-group extend and replaces with actual code

Remaining step: Replacing the `@extend .form-control;` on all input types. Any thoughts here would be welcomed.
- [] Option 1:  I'm inclined to build some custom mixins for this that we can load and then call in here as needed. 
- [] Option 2: Just copy/paste a decent amount of the code from the `_forms.scss` file in Bootstrap core....
- [] Better ideas??


src/sass/theme/_contact-form7.scss
```diff
@@ -15,60 +15,61 @@
...
+	input[type="file"],
+	input[type="submit"],
+	select,
+	textarea {
+		@extend .form-control;
+	}
 
-  // submit button, inherit .btn and .btn-outline-primary classes.
-  input[type=submit] {
-    @extend .btn;
```
IanDelMar, line 45: You remove all styles associated with .btn.

IanDelMar:
There is an open and very promising PR addressing cf7 optimization: #1467. 

Did you check whether using mixins is in fact an optimization? I doubt that this is the case for going from `@extend .btn-outline-primary;` to `@include button-outline-variant($primary);`.

I'd suggest to either only use CSS for things that cannot be easily done via the tag generator and leave adding appropriate classes to the user (one could add a cf7 template showing what should be done in order to get the desired result). See example 1 below. Or add modules for the most necessary input types (which still needs CSS for the validation stuff). See example 2 below.

### Example 1 using Bootstrap 4

```scss
.wpcf7 .screen-reader-response {

	@extend .sr-only;
}

.wpcf7 form .wpcf7-response-output {

	@extend .alert;
	@extend .alert-info;
}

.wpcf7 form.init .wpcf7-response-output,
.wpcf7 form.resetting .wpcf7-response-output,
.wpcf7 form.submitting .wpcf7-response-output {
	display: none;
}

.wpcf7 form.sent .wpcf7-response-output {

	@extend .alert-success;
}

.wpcf7 form.failed .wpcf7-response-output,
.wpcf7 form.aborted .wpcf7-response-output {

	@extend .alert-danger;
}

.wpcf7 form.spam .wpcf7-response-output {

	@extend .alert-warning;
}

.wpcf7 form.invalid .wpcf7-response-output,
.wpcf7 form.unaccepted .wpcf7-response-output,
.wpcf7 form.payment-required .wpcf7-response-output {

	@extend .alert-danger;
}

.wpcf7-not-valid {

	@extend .is-invalid;
}

.wpcf7 form.invalid .wpcf7-not-valid-tip {

	@extend .invalid-feedback;
	display: block;
}

.wpcf7-checkbox,
.wpcf7-acceptance {

	@extend .form-check;

	input {

		@extend .form-check-input;
	}
}

.wpcf7-list-item-label {

	@extend .form-check-label;
}

.wpcf7-spinner {
	visibility: hidden;

	@extend .spinner-border;
	@extend .ml-1;
}

form.submitting .wpcf7-spinner {
	visibility: visible;
}

.wpcf7-form br {
	display: none;
}
``` 

```php
// Do not load CF7 styles. We use our own styles.
add_filter( 'wpcf7_load_css', '__return_false' );

// Disallow auto paragraphs.
add_filter( 'wpcf7_autop_or_not', '__return_false', 100 );
```

```
<div class="form-group">
    <label class="sr-only" for="contact-name">Your name</label>
    [text* contact-name id:contact-name class:form-control placeholder "Your name*"]
</div>
<div class="form-group">
    <label class="sr-only" for="contact-email">Your email</label>
    [email* contact-email id:contact-email class:form-control placeholder "Your email*"]
</div>
<div class="form-group">
    <label class="sr-only" for="contact-subject">Subject</label>
    [text contact-subject id:contact-subject class:form-control placeholder "Subject"]
</div>
<div class="form-group">
    <label class="sr-only" for="contact-message">Your message</label>
    [textarea* contact-message id:contact-message class:form-control placeholder "Your message*"]
</div>
[honeypot website-743 move-inline-css:true timecheck_enabled:true "Website"]
<div class="form-group">
    [acceptance acceptance] gdpr stuff [/acceptance]
</div>
<div class="form-group">
    [response]
</div>
<div class="d-flex align-items-center">[submit class:btn class:btn-primary "Send message"]</div>
```
### Example 2 for submit
```php
add_action( 'wpcf7_init', 'understrap_add_form_tag_submit', 10, 0 );
function understrap_add_form_tag_submit() {
	wpcf7_add_form_tag( 'understrap_submit', 'understrap_submit_form_tag_handler' );
}

function understrap_submit_form_tag_handler( $tag ) {
	$class = wpcf7_form_controls_class( $tag->type, 'has-spinner' );

	$atts = array();

	$classes = $tag->get_class_option( $class );
	if ( false === strpos( $classes, 'btn' ) ) {
		$classes = 'btn ' . $classes;
	}

	$atts['class'] = $classes;
	$atts['id'] = $tag->get_id_option();
	$atts['tabindex'] = $tag->get_option( 'tabindex', 'signed_int', true );

	$value = isset( $tag->values[0] ) ? $tag->values[0] : '';

	if ( empty( $value ) ) {
		$value = __( 'Send', 'contact-form-7' );
	}

	$atts['type'] = 'submit';
	$atts['value'] = $value;

	$atts = wpcf7_format_atts( $atts );

	$html = sprintf( '<input %1$s />', $atts );

	return $html;
}

add_action( 'wpcf7_admin_init', 'understrap_add_tag_generator_submit', 55, 0 );
function understrap_add_tag_generator_submit() {
	$label = __( 'Understrap: submit', 'understrap' );
	$tag_generator = WPCF7_TagGenerator::get_instance();
	$tag_generator->add( 'understrap_submit', $label,
		'understrap_tag_generator_submit', array( 'nameless' => 1 ) );
}

function understrap_tag_generator_submit( $contact_form, $args = '' ) {
	ob_start();
	wpcf7_tag_generator_submit( $contact_form, $args );
	$boxes = ob_get_clean();
	$boxes = str_replace( 'name="submit"', 'name="understrap_submit"', $boxes );
	echo $boxes;
}
```
If you stick to adding CSS then I would provide all neccessary CSS, i.e. using the @extend method as done in the past. But as you mentioned only load the styles if the plugin is activated or only if a cf7 shortcode is present on a page. I don't really see the point going for a hybrid version where you do not add all the CSS a user might need, e.g. inline-forms.

bacoords:
I agree with all of this. There was more context on that PR in the child theme version - it was causing issues with focus selectors, so there's more work to be done if we go that route.

Personally, I would recommend we deprecate the CF7 support entirely (I understand this may not happen). It is a very popular plugin, BUT it also allows you to completely control the css/html of the form markup, making all of this additional code really unnecessary. On my local, removing CF7 takes su from 411kb to 294kb on the main stylesheet (from 356 to 242 on the minified). 

I agree that we could provide some easy documentation for adding the '.form-control' class to inputs, the `.btn-` classes to the submit. 

My second idea would be a separate CSS file that has a few bootstrap imports, enough to generate a decent stylesheet that only loads conditionally, if CF7 is activated. I'm not a huge fan of one more stylesheet in the build process, but the end result would be faster websites for many users.

danielmrey:
In case this helps, The file scr/sass/theme/_contact-form7.scss adds too many classes for the contact Form 7, instead I get a similar result with this js code that I add in src/js/custom-javascript.js:

`document.querySelectorAll('input[type="text"],input[type="email"],input[type="submit"],input[type="password"],input[type="number"],input[type="date"],input[type="tel"],textarea,select') .forEach(e => {e.className += ' form-control';});`

PR #1662, opened 2022-03-13
Adds full width landing page template

bacoords:
## Description
This adds a new landing page template- very useful for doing full designs in the block editor where you don't want the page title to show.

## Motivation and Context

Genesis uses an alternate approach (a Hide Title checkbox) but they also have a legacy page templating system to deal with. A few other themes offer this same sort of template.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I pulled my branch from `develop`.
- [x ] I am submitting my pull request to `develop`.
- [x ] I have resolved any conflicts merging this pull request would create.
- [ x] I have checked that there aren't other open Pull Requests for the same update/change.
- [x ] My code follows the code style of this project.
- [ x] I have performed a self-review of my own code.
- [ x] I have commented my code, particularly in hard-to-understand areas.
- [ x] I have read the **[CONTRIBUTING](https://github.com/understrap/understrap/blob/main/.github/CONTRIBUTING.md)** document.

Closes https://github.com/understrap/understrap/issues/1657


bacoords:
@IanDelMar The only issue I see is that there's still the wrapper padding, which adds space above and below the content. I was imaging this template being one where you could do cover blocks that start right below the header, so removing the padding. Do you think having something like this is OK?

```css
body.page-template-no-title .wrapper{ padding: 0 }
```

IanDelMar:
@bacoords If you really would like to set the padding to zero, it can be done without adding CSS:
```php
// page-templates/fullwidthpage.php

...

$no_title = is_page_template( 'page-templates/no-title.php' );
?>

<div class="<?php echo $no_title ? 'no-' : ''; ?>wrapper" id="full-width-page-wrapper">
...
```

But I don't think that this is a sustainable solution. It assumes that a cover block is the first block and that a cover block is the last block. I think having a cover block as the last block is not a common pattern and also the first block will not necessarily be a cover block. So there will be people coming up with "space is missing". I'd probably go for something like
```php
/**
 * Adds body classes to the no title page templade indicating the position
 * of cover blocks if there are any.
 *
 * @param array<int,string> $classes Array of body classes.
 * @return array<int,string> Filtered array of body classes.
 */
function no_title_page_cover_position_body_class( $classes ) {

	$wp_post = get_post();

	// Check whether it is the no title page template.
	if ( ! $wp_post instanceof WP_Post || ! is_page_template( 'page-templates/no-title.php' ) ) {
		return $classes;
	}

	// Check whether the post has a pattern of a block.
	if ( ! has_blocks( $wp_post ) ) {
		return $classes;
	}

	$blocks = parse_blocks( $wp_post->post_content );
	if ( ! is_array( $blocks ) ) {
		return $classes;
	}

	// Check whether a cover block is the first/last block. It may be both.
	if ( 'core/cover' === $blocks[0]['blockName'] ) {
		$classes[] = 'cover-first';
	}
	if ( 'core/cover' === $blocks[ count( $blocks ) - 1 ]['blockName'] ) {
		$classes[] = 'cover-last';
	}

	return $classes;
}
add_filter( 'body_class', 'no_title_page_cover_position_body_class' );
```

```scss
.page-template-no-title.cover-first .wrapper {
	padding-top: 0;
}
.page-template-no-title.cover-last .wrapper {
	padding-bottom: 0;
}
```

When the cover block is the only block on the page for logged-in users with the capability to edit pages there is the edit post link causing a gap between the block and

However, I'm not sure it makes sense to add templates when WP transforms to full site editing. But I have no idea where Understrap is heading to... Also, I think we should strive to reduce CSS and not keep adding new styles.

bacoords:
Yeah these are all good points. I'm leaving this open for discussion for a little bit so we can think on it. 

As for FSE, I've messed around with is, but since all the HTML markup HAS to be generated by Gutenberg, there's really no place for Understrap in the FSE system. (Though they may break it down and let you just edit parts like the header, the footer, etc, which would be cool and something we could support.)

I've also been considering a "headless" Understrap theme (maybe combining with a tool like Faust.js) where anyone could build a starter headless site with WordPress + Bootstrap. We'll see.

IanDelMar:
I haven't started to dive into Gutenberg. But you can create custom blocks with custom markup, no? Then that's where I see an opportunity for Understrap, namely to provide Bootstrap components as blocks.

bacoords:
At this point I'm going to merge this without messing with the padding. We can always include documentation on how to get rid of the padding for this template.

PR #1760, opened 2022-05-27
Only enqueue jQuery if Bootstrap version = 4

bbbenji:
## Description
Only enqueue jQuery if Bootstrap version = 4

## Motivation and Context
Bootstrap 5 no longer requires jQuery. This change only enqueues jQuery, if Bootstrap 4 is selected.

## Types of changes

## Checklist:

### DIFF
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -30,7 +30,9 @@ if ( ! function_exists( 'understrap_scripts' ) ) {
 		$css_version = $theme_version . '.' . filemtime( get_template_directory() . $theme_styles );
 		wp_enqueue_style( 'understrap-styles', get_template_directory_uri() . $theme_styles, array(), $css_version );
 
-		wp_enqueue_script( 'jquery' );
+		if ( 'bootstrap4' === $bootstrap_version ) {
+			wp_enqueue_script( 'jquery' );
+		}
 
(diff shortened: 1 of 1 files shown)


bacoords:
I haven't tested this out yet, but in theory this _should_ be fine. Other plugins and child themes can still load jQuery if they need it- the [starter child theme](https://github.com/understrap/understrap-child/blob/main/functions.php) does...

I'll run some tests, but open to discussion from anyone who sees any issues here, as this technically would be a change for sites that are live with BS5 and the parent theme active.

IanDelMar:
I am in favor of this change. However, this is a breaking change and I suggest to wait for other breaking changes in order to minimize the number of breaking changes.

IanDelMar:
The hero slider uses jQuery. I made a PR removing jQuery. See #1840

bacoords:
The more I think about this, the more I think it's a good thing:

- Understrap still defaults to BS4, so most users won't be affected.
- The child starter theme has separately enqueued jQuery anyway, so users of the child starter theme should be fine (though I plan to remove it there once this version is released since the child starter now only supports BS5)
- It doesn't 'remove' jQuery as much as it doesn't 'ask for it', meaning any plugin that needs it will have enabled it. If a user was adding custom JS somehow, then they should have the ability to enqueue it

Still leaving this open for comment.

PR #1836, opened 2022-07-25
Update id to match page template

IanDelMar:
For consistency, adds a page template specific ID for the wrapper in the no-title page template.

### DIFF
--- a/page-templates/fullwidthpage.php
+++ b/page-templates/fullwidthpage.php
@@ -16,9 +16,14 @@ $container = get_theme_mod( 'understrap_container_type' );
 if ( is_front_page() ) {
 	get_template_part( 'global-templates/hero' );
 }
+
+$wrapper_id = 'full-width-page-wrapper';
+if ( is_page_template( 'page-templates/no-title.php' ) ) {
+	$wrapper_id = 'no-title-page-wrapper';
+}
 ?>
 
-<div class="wrapper" id="full-width-page-wrapper">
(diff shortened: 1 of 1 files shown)


page-templates/fullwidthpage.php
```diff
@@ -16,9 +16,14 @@
...
 }
+
+$wrapper_id = 'full-width-page-wrapper';
+if ( is_page_template( 'page-templates/no-title.php' ) ) {
+	$wrapper_id = 'no-title-page-wrapper';
+}
 ?>
 
-<div class="wrapper" id="full-width-page-wrapper">
+<div class="wrapper" id="<?php echo $wrapper_id; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- ok. ?>">
```
bacoords, line 26: What's the reasoning for not just using `esc_attr()` here?

bacoords (commented):
Do we need to also match the HTML comment on line 51?

page-templates/fullwidthpage.php
IanDelMar: We know that we will never have to escape the attribute. It's either `'full-width-page-wrapper'` or `'no-title-page-wrapper'`. Calling `esc_attr()` just adds a call without any benefit.

IanDelMar:
> Do we need to also match the HTML comment on line 51?

Updated the HTML comment

bacoords (approved):

PR #1648, opened 2022-03-04
Fixed that flags do not show with WPML plugin

kirannasim:
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I pulled my branch from `develop`.
- [ ] I am submitting my pull request to `develop`.
- [ ] I have resolved any conflicts merging this pull request would create.
- [ ] I have checked that there aren't other open Pull Requests for the same update/change.
- [ ] My code follows the code style of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] \(Optional) My change requires a change to the documentation.
- [ ] \(Optional) I have updated the documentation accordingly.
- [ ] \(Optional) My change requires a change to the translations.
- [ ] \(Optional) I have updated the translations accordingly.
- [ ] `composer cs:check` has passed locally.
- [ ] `composer lint:php` has passed locally.
- [ ] I have read the **[CONTRIBUTING](https://github.com/understrap/understrap/blob/main/.github/CONTRIBUTING.md)** document.

Issue #1468, opened 2021-10-14
Compatibility with WPML - Flags do not show when language switcher in menu


nickihastings:
**Describe the problem**
I am using understrap with WPML and I am seeing a similar issue to this one https://github.com/understrap/understrap/issues/1305 with Polylang. When I activate a language switcher for a menu the image tags for the flags are being stripped out, it is also doing the same thing in the admin where it would usually show a preview.

I notice in the issue for Polylang that your last note implies that everything is working with WPML, so I am just checking whether this is something you are aware of with WPML.

**Screenshots / Videos**
![backend-flags](https://user-images.githubusercontent.com/30271858/137370600-c724b6d7-5a35-4e46-a653-2c2640bab2e7.jpg)
![frontend-flags](https://user-images.githubusercontent.com/30271858/137370612-2692e518-64e5-4587-9f41-b09e6b0cd80e.jpg)

bacoords:
This was added for the next release (v1.1.0) https://github.com/understrap/understrap/commit/7ff3bbdf8c9599f5ee14520125371ba906ab0318

bacoords:
Closed with release [1.1.0](https://github.com/understrap/understrap/releases/tag/v1.1.0).

lbriet:
1.1.0 still breaks flag display for the [Polylang](https://wordpress.org/plugins/polylang/) translation plugin. The image tag _does_ render now, but it filters out the `data:` component in the base64 encoded image src;

i.e.

the following is output

```
<img src="image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAALCAMAAABBPP0LAAAAY1BMVEX/AAD8AADzAADqAAD/fYz+dYT6aHr7XG36VGb6R1f4L0H/ECz+mKXw8fH9/f36+vr19vbW1tbS0tIAG493n89cjcZNgsA/eLkzcbUpaLAcYKwAAFEAA4ANU6UAADsAAHgAAGcSgiW+AAAAS0lEQVR4AQXBiQ3CQBAAMc/dConQf688SoIdBIIyQiR9R0tCuM2rNxHpMqsDIWiBEE4NgnBiHhVJcs48P0uSjXsPl4hLmH2EHzb4A7DPDGppZMSwAAAAAElFTkSuQmCC" alt="Nederlands" width="16" height="11" style="width: 16px;height: 11px">
```

instead of

```
[...]

Issue #1305, opened 2021-07-07
The language flags of the menu items of the Polylang plugin are not visible.


ramedey:
## Issue is about..
  - [X] a Bug Report
  - [X] I need help

## Please tell us about your environment:

  - **WordPress Version:** 5.7.2
  - **Polylang Version**: 3.0.6
  - **UnderStrap Version**: 0.9.6 

---

# Issue Overview for Bugreports

The language flags of the menu items of Polylang are not visible. The menu item HTML markup is there but the image element is not.

## Steps to reproduce?
This is what I did to reproduce it:

1- Create a new fresh Wordpress site
2- Install Understrap theme
3- Install Polylang and add English and Dutch
4- Add translation to the sample page
5- Add language switcher to the top menu (with the language flag option)
6- Check the result in the Twenty Twenty One theme. The flag image is visible, everything is fine.
7- Switch to Understrap theme
8- Check the website again. This time the flag is not visible.

Could you please help?

Thank you for considering my request.

bacoords:
Do you have this on a live site we can look at?

ramedey:
I have created a live site that reproduces the issue:

Url: https://testunderstrap.000webhostapp.com/
user: admin
password: testunderstrap

The active theme is Twenty Twenty-One. You will see the flag in the top menu, and you can click and switch from English to Dutch. Everything works as expected.

The Understrap theme is installed, I'd suggest you activate it to se what happens. FYI, I didn't modify any file.

When you switch to the Understrap theme you will notice that the menu item is still there (you can hover over it or just check the HTML), but the actual flag image is gone.

Please let me know if you have any more questions.

Thanks for your help.

bacoords:
I was able to track down the issue to this filter right here: 
https://github.com/understrap/understrap/blob/main/inc/extras.php#L159
```php
// Escapes all occurances of 'the_title()' and 'get_the_title()'.
add_filter( 'the_title', 'understrap_kses_title' );
```

Essentially we're stripping some HTML from the_title, which is a filter that also happens to run when building the HTML of a  navigation menu. This causes the flag image icons to be removed. You can probably fix this by either:

1 - Removing the filter in a child theme.

2 - Overwriting the function with something like this (untested) in a child theme:

```php
function understrap_kses_title( $title ){
    return $title;
}
```

3 - Or you can use the filter `understrap_kses_title` to add image tags to the array of allowed HTML.

[...]

PR #1850, opened 2022-08-04
Replace understrap_kses_title() with wp_kses_post()

IanDelMar:
## Description
This PR deprecates the `understrap_kses_title()` filter and replaces it with `wp_kses_post()`.

## Motivation and Context
`understrap_kses_title()` causes issues by aggressively stripping HTML from titles. To escape titles with `wp_kses_post()` is less strict but still strips scripts from titles.

## Types of changes

## Checklist:

#1305
#1468
#1648
#1849

### DIFF
--- a/inc/deprecated.php
+++ b/inc/deprecated.php
@@ -8,6 +8,35 @@
 // Exit if accessed directly.
 defined( 'ABSPATH' ) || exit;
 
+if ( ! function_exists( 'understrap_kses_title' ) ) {
+	/*
+	 * Throw a deprecation notice if the function is not defined but used by
+	 * the child theme.
+     */
+
+	/**
+	 * Sanitizes data for allowed HTML tags for post title.
+	 *
+	 * @param string $data Post title to filter.
+	 * @return string Filtered post title with allowed HTML tags and attributes intact.
+	 *
+	 * @deprecated 1.2.0 Replaced with `wp_kses_post()`.
+	 */
+	function understrap_kses_title( $data ) {
+
+		_deprecated_function( 'understrap_kses_title', '1.2.0' );
+
+		global $allowedposttags;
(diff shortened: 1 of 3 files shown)


IanDelMar:
@bacoords ready for review

bacoords:
@IanDelMar This is a much better approach. When you have time can you fix the conflicts from me pulling in the other PRs?

IanDelMar:
@bacoords done

bacoords:
I was hoping this would fix the issue with Polylang using images in the nav menu items (https://github.com/understrap/understrap/issues/1305) like it does with WPML (https://github.com/understrap/understrap/issues/1468) but the images still don't work. This is because they're using data URIs for the flag icons.

My assumption is that we'd need to re-open that old issue if we went this route (I still like this approach) and then either 
add the 'data' protocol to [`kses_allowed_protocols`](https://developer.wordpress.org/reference/hooks/kses_allowed_protocols/) or else add some documentation explaining how to do it for Polylang users.

Or we need to take step back and decide why we're even sanitizing titles to begin with and if this is a common practice.

IanDelMar:
> Or we need to take step back and decide why we're even sanitizing titles to begin with and if this is a common practice.
>

I don't think that it is common practice. The reason I added this was that titles are not sanitized and, e.g., a malicious plugin can add a script. Probably this was too ambitious. We do not sanitize `the_content()` either. In terms of the original purpose it may be sufficient to just strip script tags and onclick attributes? At the end of the day, you can't force the title or the content to be "scriptless" via a filter unless it's priority is infinitely large. You can only make sure that a script already stored in the database will not be executed.

bacoords:
@IanDelMar Closing this in favor of https://github.com/understrap/understrap/pull/1927

PR #1935, opened 2022-10-03
Update footer widget role to "complementary"

### LINKED ISSUE #1775: Footer Full widget area role="footer" markup causes "[role] values are not valid" in Lighthouse testing
### The Bug

**Describe the bug**

In sidebar-footerfull.php, this ARIA role markup is causing a lower accessibility score in Lighthouse:
`<div class="wrapper" id="wrapper-footer-full" role="footer">`

It reports the error: **[role] values are not valid**
and provides a link to [more information](https://www.w3.org/TR/wai-aria-1.1/#role_definitions), showing that 'footer' is not a valid ARIA role definition.

**To Reproduce**
1. Activate the Footer Full widget area in Understrap
2. In Chrome, test an Understrap page with Lighthouse
3. Look in the Lighthouse accessibility report

**Expected behavior**
Footer Full markup should use a valid ARIA role, improving accessibility and Lighthouse scores.

### The Environment

- **WordPress Version:** 6.0.0
- **Browser:** Chrome Version 102.0.5005.115 (Official Build) (x86_64)  
- **Language:** PHP / HTML

bacoords:
Fix for issue with `role="footer"` in lighthouse. Updates the role to "complementary" (also considering changing the element to an aside, but that could be more of a breaking change).

Closes #1775

### DIFF
--- a/sidebar-templates/sidebar-footerfull.php
+++ b/sidebar-templates/sidebar-footerfull.php
@@ -16,7 +16,7 @@ $container = get_theme_mod( 'understrap_container_type' );
 
 	<!-- ******************* The Footer Full-width Widget Area ******************* -->
 
-	<div class="wrapper" id="wrapper-footer-full" role="footer">
+	<div class="wrapper" id="wrapper-footer-full" role="complementary">
 
 		<div class="<?php echo esc_attr( $container ); ?>" id="footer-full-content" tabindex="-1">
 


howarddc (approved):

sidebar-templates/sidebar-footerfull.php
```diff
@@ -16,7 +16,7 @@
 
 	<!-- ******************* The Footer Full-width Widget Area ******************* -->
 
-	<div class="wrapper" id="wrapper-footer-full" role="footer">
+	<div class="wrapper" id="wrapper-footer-full" role="complementary">
```
IanDelMar, line 19:
```suggestion
	<aside class="wrapper" id="wrapper-footer-full">
```
And 
```diff
-</div><!-- #wrapper-footer-full -->
+</aside><!-- #wrapper-footer-full -->
```
on [line 31](https://github.com/understrap/understrap/blob/7b7fe079c0d60873c980aa67bf12e88105a49f81/sidebar-templates/sidebar-footerfull.php#L31)

bacoords: My concern here was about changing the semantic markup - I ran some tests and haven't seen any issues changing the `div` to an `aside` - how about you?

IanDelMar: Sorry, I was on a hurry, remembered the issue, saw your changes and didn't check your comment. You're right, a markup change should not be put in a feature release. Maybe it's best to open a different PR with milestone 2.0 so we don't forget to change the `div` to an `aside`?

PR #1998, opened 2022-11-02
Update Woocommerce templates

IanDelMar:
## Description
This PR updates the version number in Woocommerce templates.

## Motivation and Context
WooCommerce 7.0.1 adds `wc_wp_theme_get_element_class_name()` which is a wrapper for `wp_theme_get_element_class_name()` which calls `WP_Theme_JSON::get_element_class_name()`. Understrap does not come with a theme.json. Therefor a call to `WP_Theme_JSON::get_element_class_name( $element )` gives an empty string and adding `wc_wp_theme_get_element_class_name()` to the templates can be skipped (would be a breaking change).

## Types of changes

#1997

### DIFF
--- a/woocommerce/cart/cart-empty.php
+++ b/woocommerce/cart/cart-empty.php
@@ -12,7 +12,7 @@
  *
  * @see     https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 3.5.0
+ * @version 7.0.1
  */
 
 defined( 'ABSPATH' ) || exit;
--- a/woocommerce/cart/cart.php
+++ b/woocommerce/cart/cart.php
@@ -12,7 +12,7 @@
  *
  * @see     https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 3.8.0
+ * @version 7.0.1
  */
 
 defined( 'ABSPATH' ) || exit;
--- a/woocommerce/cart/proceed-to-checkout-button.php
+++ b/woocommerce/cart/proceed-to-checkout-button.php
@@ -14,7 +14,7 @@
  *
  * @see     https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
- * @version 2.4.0
+ * @version 7.0.1
  */
 
 // Exit if accessed directly.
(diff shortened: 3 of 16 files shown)


bacoords:
@IanDelMar I'd like to fast-track this to do another minor release. I just need to spend some time testing. Let me know what you think.

IanDelMar:
Think about what? Releasing 1.3.0 soon? Actually this doesn't qualify for a feature release, does it? There is no code change at all. It is a patch (1.2.2) at max, no?

bacoords:
@IanDelMar Exactly! - I already updated the tag to 1.2.2 earlier

And yeah the readme makes sense

IanDelMar:
Oh, didn't notice the updated version! And yes, I think Understrap+WooCommerce users will be happy if you release 1.2.2 soon. 

Concerning the supported version: the maximum version is clear. Finding the minimum version is a little cumbersome. I tend to  simply making a cut and say v6+ is supported (currently . I'm not sure what that means for the type of change. Is it a breaking change which I guess?

Active installs of WooCommerce as of today: versions from 6.8 account for 45.9% of installs.

image
[Source](https://wordpress.org/plugins/woocommerce/advanced/)

IanDelMar:
For 1.2.2: We should switch back to inc/editor-color-palette.json with Bootstrap 4 colors.

bacoords:
@IanDelMar Looks like we're just missing the 7.1.0 update to `global/form-login.php`. From what I can gather, they're adding the `woocommerce-Input` class to the password field. It's unclear if it's a typo and they meant 7.0.1, but either way the class is missing.

Re: `wc_wp_theme_get_element_class_name()` - I think we should still include it. I often add a theme.json to my child themes, even if we couldn't really add one to the parent Understrap theme without breaking a million sites. It seems better to include the function for users who might have a theme.json, than just not include it. https://github.com/woocommerce/woocommerce/commits/trunk/plugins/woocommerce/templates/global/form-login.php

IanDelMar:
@bacoords then we have to provide both `wc_wp_theme_get_element_class_name()` and `wp_theme_get_element_class_name()`

Simply adding `wc_wp_theme_get_element_class_name()` would break sites on WooCommerce v6 or WP < 6.1.0.

PR #2111, opened 2023-06-23
Fix typos

szepeviktor:
Tool used: https://github.com/crate-ci/typos
Configuration `.typos.toml`
```toml
[files]
extend-exclude = [
  "css/",
  "fonts/",
  "js/*.min.js",
  "js/*.map",
  "js/theme*.js",
  "languages/",
  "src/sass/assets/bootstrap4/",
  "src/sass/assets/bootstrap5/",
  "src/sass/assets/fontawesome/"
]
[default.extend-words]
optin = "optin"
adn = "adn"
```

[_source_](https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery-spelling.yml)

PR #2113, opened 2023-07-04
Fix typos

szepeviktor:
From #2111 
@bacoords Everything is done 🍏


szepeviktor:
Please consider using a package manager and a build process to get dependencies like Bootstrap.

inc/class-wp-bootstrap-navwalker.php
```diff
@@ -407,7 +407,7 @@ public static function fallback( $args ) {
 		 *
 		 * @return array  $classes         a maybe modified array of classnames.
 		 */
-		private function seporate_linkmods_and_icons_from_classes( $classes, &$linkmod_classes, &$icon_classes, $depth ) {
+		private function separate_linkmods_and_icons_from_classes( $classes, &$linkmod_classes, &$icon_classes, $depth ) {
```
bacoords, line 410: I think this is the only change I'd like to see removed from this PR. If we want to update the name of this function, I think we'd need to still keep the older one and deprecate it to be safe, but I think we'd want that in a separate PR.

bacoords (changes requested):
Looks great! One change request, though open to feedback from others (CC @IanDelMar )

inc/class-wp-bootstrap-navwalker.php
bacoords: Just realizing we already had this discussion about this being a private function. So to be clear - we see no reason to add the old function with a deprecation notice, correct?

szepeviktor: Yes, correct.

szepeviktor:
@bacoords Included that new finding in this PR.

inc/class-wp-bootstrap-navwalker.php
pattonwebz: Noting here that once this is merged to understrap I would be more than happy to merge the same changeset to the navwalker repo as well. I have partially stepped away from that project, but I trust that the changes here are being tested and done for the best of reasons so would want to roll that back to the origin repo as well :)

bacoords (approved):

PR #2121, opened 2023-10-09
Adds BS container classes and max-widths to editor styles

### LINKED ISSUE #2083: Feature Request: Better support of Bootstrap inside the block editor
We've been slowly adding some bootstrap features into the block editor (which has gotten easier over time as the block editor's classes get more scoped and stable). This issue will be a checklist of things I've been doing in my custom child themes to make the `src/sass/custom-editor-style.scss` file more useful. The goal would be to bring these into understrap.

- [ ] Importing the bootstrap grid / containers
- [ ] Mapping the block width to the bootstrap containers

## Problem Statement

The block editor still doesn't quite match the front-end

## Who will benefit?

Anyone using the block editor but especially those using Bootstrap grid classes in their columns.

## Benefits and risks

There'll probably be a lot of compatibility checking. For example, anyone using the container-fluid globally will still not see the right layouts (but it will be closer).

## Proposed solution

`src/sass/custom-editor-style.scss`:

**Importing the bootstrap grid / containers**
This allows for more integration in custom/acf blocks.
```scss
@import "assets/bootstrap5/containers";
@import "assets/bootstrap5/grid";
```

**Mapping the block width to the bootstrap containers:**
Because blocks don't get the max-width on the front-end that Gutenberg provides on the backend (and because there's no way to tell Gutenberg that you have multiple max-widths) this is a decent middle ground that better approximates the front end
```scss
@each $breakpoint, $container-max-width in $container-max-widths {
    @include media-breakpoint-up($breakpoint, $grid-breakpoints) {
		html :where(.wp-block) {
			max-width: $container-max-width;
		}
	}
}
```

## Priority/Severity

- [ ] High (This will bring a huge increase in performance/productivity/usability/legislative cover)
- [x] Medium (This will bring a good increase in performance/productivity/usability)
- [ ] Low (anything else e.g., trivial, minor improvements)

bacoords:
## Description
This PR imports the grid/container classes into the block editor for development teams that are using things like `.container` in their custom blocks or ACF blocks. 

This also maps the default container widths to the blocks using the same `where(.wp-block)` selector that core uses. So instead of all of the blocks having a different content width on the back and front ends, the block editor looks a lot closer to the frontend of the site.

## Motivation and Context
Closes #2083

## Types of changes

- [ ] Refactoring (a code change that neither fixes a bug nor adds a feature)
- [ ] Styling (changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc))
- [ ] Development (changes that affect the build system or external dependencies)
- [ ] Documentation (changes that affect existing inline documentation)

## Checklist:

- [ ] `composer phpcs` has passed locally.
- [ ] `composer php-lint` has passed locally.
- [ ] `composer phpmd` has passed locally.
- [ ] `composer phpstan` has passed locally.


IanDelMar:
`html :where(.wp-block) {` won't work in older WP versions that still support IE. It's also not supported by all browsers that Understrap claims to support for Bootstrap 4:
https://github.com/understrap/understrap/blob/546b350b62c22d384a563a9c0ad1b77f2331e046/.browserslistrc#L12-L28

image
Source: https://caniuse.com/mdn-css_selectors_where

bacoords:
Interesting. IE11 support was dropped in WP in 5.8. Currently we recommend 5.0 as the minimum version for this theme. According to [stats](https://wordpress.org/about/stats/) about 82% of WP sites are at 5.8 or above. 

It's also worth noting that Bootstrap 5 does not support IE, so technically theme does support it but only if the user doesn't switch to BS5.

So the options would be
- intentionally declare a higher supported version (5.8)
- rewrite the CSS (in the BS4 version) to be IE friendly but potentially different than how the block editor does it. This might require more testing since [`:where`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) intentionally has a lower specificity and is easier to override (for example when the block editor inspector control is showing)

Issue #2159, opened 2023-12-04
Feature Request: Add `.nvmrc` file with the needed version of node/npm


pattonwebz:
## Problem Statement
Some versions of node/npm cause problems with installing and running specific build scripts. With a way to tell users what version is needed, they will avoid running into these issues.

## Who will benefit?
Anyone that uses nvm to get the correct version of node/npm for projects.

## Benefits and risks
There are no risks here as the `.nvmrc` file is benign and has no particular effect on the code. The benefit is users would have a way to know and use the version of node/npm that is required by the project.

## Proposed solution
Add a `.nvmrc` file to the root of the project once we know the exact version that is needed.

## Priority/Severity

- [ ] High (This will bring a huge increase in performance/productivity/usability/legislative cover)
- [x] Medium (This will bring a good increase in performance/productivity/usability)
- [ ] Low (anything else e.g., trivial, minor improvements)

pattonwebz:
I can work on this sometime this week if someone can confirm an exact version of node that is to be used.

bacoords:
In the `package.json` we reference `>=16` so I'm guessing that'd be https://nodejs.org/en/blog/release/v16.20.2

I can test it out as well.

pattonwebz:
If we use `>=16` in the package file then I think we can safely put `v16` in the .nvmrc and let the system decide what specific version to use. Will open a PR now.

PR #2160, opened 2023-12-04
Create .nvmrc file pointing to `v16`

pattonwebz:
Adds an `.nvmrc` file that sets `v16` as the node/npm version of choice.

Users with `nvm` can shift to the expected version by running `nvm use` in their terminal.

Closes: #2159


bacoords (dismissed):

bacoords (approved):
(Re-approving after rebasing the PR to `develop`)

PR #2209, opened 2024-04-01
Remove past entry about dependency updates

IanDelMar:
See https://github.com/understrap/understrap/actions/runs/8515166522/job/23322131310

While this is a false positive, I see no reason to have dependency updates in the change log, especially if they do not change the build process.


IanDelMar:
@bacoords This is safe to merge.

